PR #17, opened 2021-04-26
Staging

moritzhaller:


seed/serialize.py
```diff
@@ -125,6 +125,9 @@ def serialize_and_save_variations_seed_message(seed_data, path):
         if 'min_version' in study_data['filter']:
             study.filter.min_version = study_data['filter']['min_version']
 
+        if 'max_version' in study_data['filter']:
```
iefremov, line 128: i don't quite follow why github shows this diff, because it should be already in place

iefremov (approved):

seed/serialize.py
moritzhaller: indeed that's odd. I see we're merging 5 commits, maybe that's just how the diffs play out.

PR #37, opened 2021-05-18
Add Brave Ads custom ad notification feature study

tmancey:
60% of users should see custom and 40% of users should see system ad notifications

Custom ad notifications should be disabled on the beta and release channels

Windows (if custom ad notifications are shown):
50% of users should see custom ad notifications positioned top-right of the screen
50% of users should see custom ad notifications positioned bottom-right of the screen, to the left of system notifications (default screen position)

macOS (if custom ad notifications are shown):
50% of users should see custom ad notifications positioned bottom-right of the screen
50% of users should see custom ad notifications positioned top-right of the screen, to the left of system notifications (default screen position)


seed/seed.json
```diff
@@ -134,6 +134,122 @@
...
+                        "disable_feature": []
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 60
+                }
+            ],
+            "filter": {
+                "channel": ["UNKNOWN", "NIGHTLY"],
```
goodov, line 188: do we have some undocumented use for `UNKNOWN`?

looking at https://github.com/brave/brave-variations/blob/main/seed/proto/study.proto#L205-L207
it shouldn't be used in the seed file.
```
    // UNKNOWN value is defined here for the benefit of code using this enum
    // type, but is not actually meant to be encoded in the protobuf.
    UNKNOWN = -1;
```

seed/seed.json
```diff
@@ -134,6 +134,122 @@
...
+                    "probability_weight": 100,
+                    "parameters": [
+                        {
+                            "name": "should_show_custom_notifications",
+                            "value": "false"
+                        }
+                    ],
+                    "feature_association": {
+                        "enable_feature": ["AdNotifications"],
+                        "disable_feature": []
```
goodov, line 152: `AdNotificationsDisabledOnBetaAndReleaseChannelsStudy`
should the feature be disabled? the name and the actual result doesn't match.

seed/seed.json
```diff
@@ -134,6 +134,122 @@
...
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 60
```
goodov, line 184: I see `AdNotifications` is [enabled by default](https://github.com/brave/brave-core/blob/6cda51428c6a1a8268d3dcdf4485ca9cdb9601af/components/brave_ads/browser/features.cc#L13-L14). This is dangerous in general for new features, and also this PR doesn't do what you want it to do: it doesn't disable the feature for 60%.

You either need to explicitly disable the feature here, or disable it in code and enable it using variations (preferred way).
Explicit disable for default-enabled feature is not the best solution, because variations applied only after restart, it means that the first run a feature will always be enabled.

seed/seed.json
moritzhaller: @goodov `UKNOWN` targets dev channel. Dev builds should use CLI overrides for testing, not sure if there are other dev channel browsers in the wilde that we need to target @tmancey?

goodov: got it, but the channel can be overridden using CLI, so I'm not sure why we should do this in stage/prod branches at all.

moritzhaller: good point, I probably wrote the script before I knew about the CLI args so if @tmancey or others don't have any other reasons for supporting it we can remove `UNKNOWN` from supported channels

seed/seed.json
moritzhaller: my understanding is that this is an old feature that just received a feature flag, so somewhat unusual as we're just introducing features to ads. In this case the group split should be controlled by `should_show_custom_notifications`, where old/default behaviour is "should not show" and the configuration under test is "should show". Does that make sense?

moritzhaller: the study should probably be called `BraveAds.CustomAdNotificationsEnabledOnUnknownAndNightlyChannelStudy`, wdyt @tmancey?

seed/seed.json
moritzhaller: @goodov same as below, this is due to the unusual case of having introduced the feature only now. The feature should be enabled for all groups, the group split is along the parameter `should_show_custom_notifications`.

tmancey: I will rename now. Thanks

seed/seed.json
tmancey: Agreed. Thanks

tmancey: Fixed

seed/seed.json
tmancey: Fixed

seed/seed.json
tmancey: Removed `UNKNOWN`

seed/seed.json
goodov: > my understanding is that this is an old feature that just received a feature flag, so somewhat unusual as we're just introducing features to ads. In this case the group split should be controlled by should_show_custom_notifications, where old/default behaviour is "should not show" and the configuration under test is "should show". Does that make sense?

yep, it makes sense now, but very unusual indeed. thanks!

moritzhaller (approved):
++

goodov (approved):

PR #40, opened 2021-05-24
Increase maximum ad notifications per day for tier 1 and 2 regions

tmancey:


seed/seed.json
```diff
@@ -323,6 +323,34 @@
                 "channel": ["NIGHTLY"],
                 "platform": ["WINDOWS", "MAC", "LINUX"]
             }
+        },
+        {
+            "name": "AdServingIncreaseMaximumAdNotificationsPerDayTo60ForTier1And2Regions",
+            "experiments": [
+                {
+                    "name": "AdServing",
```
iefremov, line 331: I'd suggest we use "Enabled" or something similar as the experiment name

this is how it looks on brave://version
```
DoNotShowCustomAdNotificationsOnBetaAndReleaseChannelsStudy:AdNotifications
EphemeralStorageStudy:Enabled
PermissionLifetimeStudy:Enabled
```

seed/seed.json
```diff
@@ -323,6 +323,34 @@
                 "channel": ["NIGHTLY"],
                 "platform": ["WINDOWS", "MAC", "LINUX"]
             }
+        },
+        {
+            "name": "AdServingIncreaseMaximumAdNotificationsPerDayTo60ForTier1And2Regions",
```
iefremov, line 328: Does it make sense to specify `60` in the study name? We could modify the parameter in the future. 
maybe just `MaxAdNotificationsPerDayStudy`?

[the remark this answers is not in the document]
moritzhaller: confirmed that this is not needed as discussed with @tmancey

seed/seed.json
moritzhaller: fixed as requested by @tmancey

seed/seed.json
moritzhaller: fixed as requested by @tmancey

iefremov (approved):

PR #41, opened 2021-05-25
Update Brave Ads studies

tmancey:
See comments for changes. Also added "Disabled" node for consistency with other studies

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -139,7 +139,7 @@
             "name": "DoNotShowCustomAdNotificationsOnBetaAndReleaseChannelsStudy",
             "experiments": [
                 {
-                    "name": "AdNotifications",
+                    "name": "Enabled",
                     "probability_weight": 100,
                     "parameters": [
                         {
@@ -148,8 +148,14 @@
                         }
                     ],
                     "feature_association": {
-                        "enable_feature": ["AdNotifications"],
-                        "disable_feature": []
+                        "enable_feature": ["AdNotifications"]
+                    }
+                },
+                {
+                    "name": "Disabled",
+                    "probability_weight": 0,
+                    "feature_association": {
+                        "disable_feature": ["AdNotifications"]
                     }
                 },
                 {
@@ -167,7 +173,7 @@
             "name": "BraveAds.DoNotShowCustomAdNotificationsOnNightlyChannelStudy",
             "experiments": [
                 {
-                    "name": "AdNotifications",
+                    "name": "Enabled",
                     "probability_weight": 40,
                     "parameters": [
                         {
@@ -176,8 +182,14 @@
                         }
                     ],
                     "feature_association": {
-                        "enable_feature": ["AdNotifications"],
-                        "disable_feature": []
+                        "enable_feature": ["AdNotifications"]
+                    }
+                },
+                {
+                    "name": "Disabled",
+                    "probability_weight": 0,
+                    "feature_association": {
+                        "disable_feature": ["AdNotifications"]
                     }
                 },
                 {
@@ -192,10 +204,10 @@
             }
         },
         {
-            "name": "BraveAds.ShowCustomAdNotificationsOnNightlyChannelStudy",
+            "name": "BraveAds.ShowCustomAdNotificationStudy",
             "experiments": [
                 {
-                    "name": "AdNotifications",
+                    "name": "Enabled",
                     "probability_weight": 60,
                     "parameters": [
                         {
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -192,10 +204,44 @@
...
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 40
+                }
+            ],
+            "filter": {
+                "min_version": "91.1.26.37",
+                "channel": ["NIGHTLY", "BETA"],
```
tmancey, line 236: Show custom ad notifications to 60% of users on Windows, macOS and Linux for both Beta and Nightly build channels

seed/seed.json
```diff
@@ -216,14 +268,14 @@
             "filter": {
                 "min_version": "91.1.26.37",
                 "channel": ["NIGHTLY"],
-                "platform": ["WINDOWS", "MAC", "LINUX", "ANDROID"]
+                "platform": ["ANDROID"]
```
tmancey, line 271: Show custom ad notifications to 60% of users on Android for Nightly build channels

seed/seed.json
```diff
@@ -254,15 +312,15 @@
                 }
             ],
             "filter": {
-                "channel": ["NIGHTLY"],
+                "channel": ["NIGHTLY", "BETA"],
```
tmancey, line 315: Add beta build channel to study

seed/seed.json
```diff
@@ -293,15 +357,15 @@
...
```
tmancey, line 360: Add beta build channel to study

seed/seed.json
```diff
@@ -320,7 +390,7 @@
...
```
tmancey, line 393: Add beta build channel to study

seed/seed.json
```diff
@@ -310,8 +374,14 @@
                         }
                     ],
                     "feature_association": {
-                        "enable_feature": ["AdNotifications"],
-                        "disable_feature": []
+                        "enable_feature": ["UserActivity"]
```
tmancey, line 377: Fix bug where we had the wrong feature name for "idle_time_threshold"

seed/seed.json
```diff
@@ -139,7 +139,7 @@
             "name": "DoNotShowCustomAdNotificationsOnBetaAndReleaseChannelsStudy",
```
moritzhaller, line 139: as this is not tagged to the confirmations DTO and the default for `should_show_custom_notifications` is `false`, should we maybe remove this study to keep things simple? https://github.com/brave/brave-core/blob/24fe16e3e30a7e75cf81f16bcbaa2d2982af17f0/components/brave_ads/browser/features.cc#L99

seed/seed.json
```diff
@@ -167,7 +173,7 @@
             "name": "BraveAds.DoNotShowCustomAdNotificationsOnNightlyChannelStudy",
```
moritzhaller, line 173: same as above, as this is the default this study seems to do nothing

seed/seed.json
```diff
@@ -216,14 +268,14 @@
...
+                "platform": ["ANDROID"]
             }
         },
         {
             "name": "BraveAds.CustomAdNotificationPositionedBottomRightOffsetToTheLeftOfSystemNotificationsOnWindowsStudy",
             "experiments": [
                 {
-                    "name": "AdNotifications",
+                    "name": "Enabled",
                     "probability_weight": 50,
```
moritzhaller, line 279: this group won't be 50% of `BraveAds.ShowCustomAdNotificationStudy::Enabled` windows users, but less as clients sign up for this study independently from the custom notification study. Maybe @goodov or @iefremov can confirm. To fix this and for better interpretability, would it make sense to re-organise all ads notifications studies by platform, i.e. have one `BraveAds.ShowCustomAdNotificationStudy` for each of the following filters:

1. Windows on Nightly and Beta: 30% custom/position-1, 30% custom/position-2, 40% default
2. Mac on Nightly and Beta: 30% custom/position-1, 30% custom/position-2, 40% default
3. Linux on Nightly and Beta: 60% custom, 40% default
4. Android on Nightly: 60% custom, 40% default

as mentioned above, we don't need the "do not show" studies as this is the default in https://github.com/brave/brave-core/blob/24fe16e3e30a7e75cf81f16bcbaa2d2982af17f0/components/brave_ads/browser/features.cc#L20

tmancey: This study will only affect users who have custom ad notifications enabled and will mean 50% of users ### will see custom ad notifications bottom right offset to the left of system notifications and the other 50% will see custom ad notifications top right (default).

tmancey: With regards to the "do not show" studies, we will need these studies for backwards compatibility before the change in https://github.com/brave/brave-core/blob/24fe16e3e30a7e75cf81f16bcbaa2d2982af17f0/components/brave_ads/browser/features.cc#L20

tmancey: I do not see how the above filters would work as we need "min_version" and "max_version" for backwards compatibility.

seed/seed.json
tmancey: These studies have `max_version` and are for backwards compatibility

seed/seed.json
tmancey: These studies have `max_version` and are for backwards compatibility

PR #45, opened 2021-05-26
Add Brave Ads user activity study

tmancey:


seed/seed.json
```diff
@@ -397,6 +397,44 @@
...
+                    "name": "UserActivityTriggers.None",
+                    "probability_weight": 100,
+                    "parameters": [
+                        {
+                            "name": "triggers",
+                            "value": ""
+                        }
+                        {
+                            "name": "threshold",
+                            "value": "0.0"
```
goodov, line 413: are you sure this will work properly?
https://github.com/brave/brave-core/blob/ff5e1645a3c42fb15afa7420d669ee5fc13abc44/vendor/bat-native-ads/src/bat/ads/internal/user_activity/user_activity_scoring_util.cc#L27-L30

`triggers.empty() == true` means `score == 0`
`score < threshold? nope.`
`return true.`

tmancey: Returning true is correct as the user does not need to trigger any events to be considered active. Thanks

moritzhaller (approved):

PR #71, opened 2021-06-25
Enable native cosmetic filtering backend

antonok-edm:


seed/seed.json
```diff
@@ -1,6 +1,34 @@
...
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 0
+                }
+            ],
+            "filter": {
+                "min_version": "91.0.4472.114",
+                "channel": ["RELEASE"],
```
pes10k, line 28: Why not all channels?

antonok-edm: Good call, updated

pes10k (approved):

moritzhaller (approved):

PR #76, opened 2021-06-28
Enable native cosmetic filtering backend on production

antonok-edm:
Verified that this appears correctly on https://griffin.brave.com and in brave://version on Linux/Release when using the staging URL.

PR #81, opened 2021-07-06
Enable native cosmetic filtering on beta channel

antonok-edm:
Followup to #76, which was merged a week ago. I haven't seen any regressions reported so far.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -25,7 +25,7 @@
             ],
             "filter": {
                 "min_version": "91.0.4472.114",
-                "channel": ["NIGHTLY"],
+                "channel": ["NIGHTLY", "BETA"],
                 "platform": ["WINDOWS", "MAC", "LINUX"]
             }
         },


kjozwiak:
I'll wait till either @moritzhaller or @bridiver reviews the code before merging 👍

kjozwiak (approved):

seed/seed.json
```diff
@@ -25,7 +25,7 @@
             ],
             "filter": {
                 "min_version": "91.0.4472.114",
```
iefremov, line 27: this is wrong version. it should be `<chrome>.<brave>`, see other entries @antonok-edm @kjozwiak

iefremov: @moritzhaller maybe we could add a version validation step?

antonok-edm: ah, so it's the major version of Chromium followed by all 3 version sections of Brave?

At a first glance it looked just like a full Chromium version, so that's what I used. Probably would be good to add this information to the README at the very least.

PR #87, opened 2021-08-04
Add and enable `BraveTalk` feature on Nightly

bsclifton:
Enable feature on NIGHTLY only which was introduced with https://github.com/brave/brave-core/pull/9605

For more info about rollout, please see https://github.com/brave/brave-together/issues/112

cc: @kjozwiak

PR #89, opened 2021-08-04
Add and enable `BraveTalk` feature on Nightly (production)

bsclifton:
Production uplift of https://github.com/brave/brave-variations/pull/87


kjozwiak (approved):
Uplift into **production** approved after deliberating with @brave/uplift-approvers. QA has verified that `BraveTalkStudy` is working correctly on staging as per https://github.com/brave/brave-core/pull/9605#issuecomment-892905012.

PR #94, opened 2021-08-12
Increase inline content ads per hour and per day

tmancey:
Increase inline content ads per hour to 8 and per day to 40

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -579,12 +579,20 @@
             "name": "AdServingStudy",
             "experiments": [
                 {
-                    "name": "MaximumAdNotificationsPerDay=100",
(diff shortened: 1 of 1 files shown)


moritzhaller (approved):

seed/seed.json
```diff
@@ -579,12 +579,20 @@
             "name": "AdServingStudy",
             "experiments": [
                 {
-                    "name": "MaximumAdNotificationsPerDay=100",
+                    "name": "MaximumAdNotificationsPerDay=100/MaximumInlineContentAdsPerHour=8/MaximumInlineContentAdsPerDay=40",
```
iefremov, line 582: not sure this is convenient, mb use shorter names?..

iefremov (approved):

seed/seed.json
tmancey: We have been using the same name as the related params

PR #106, opened 2021-08-24
bump speedreader to 15%

keur:


kjozwiak (approved):

PR #108, opened 2021-08-30
update regex for ConversionsParametersStudy

iambrianfung:
# Description
Update regex on Griffin to support spec`[A-Za-z0-9\-]` in the meta-tag for verifiable ad conversions.
This is a temporary solution until https://github.com/brave/brave-browser/issues/17039

# Testing Instructions 

Download and unzip file:
[meta.html.zip](https://github.com/brave/brave-variations/files/7078382/meta.html.zip)

1. Use Charles to map meta.html to `https://brave.com?vac=variation`
2. Confirm that you see "meta tag test page" on the page for successful mapping

3. Load fresh profile with **staging** catalog
4. Enable Ads (geo-US)
5. Force creativeInstanceID `de8274a9-716b-4340-afa1-9bfa52289077` 
    - Option a. Open up local db, `creative_ads` table, and remove all other creatives
    -  Option b. make local catalog and map local via Charles
6. Trigger Push Notification Ad
7. Go to conversion page `https://brave.com?vac=variation`
8. Open local db, `conversions_queue` table, confirm that extracted `conversion_id`. 
    - expected value is `Griffin-9876-meta`
9. Force conversion confirmation to send.
10. Confirm payload contains `conversionEnvelope`

PR #115, opened 2021-09-01
add ConversionsParametersStudy

iambrianfung:
production uplift of https://github.com/brave/brave-variations/pull/108

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -831,6 +831,32 @@
             }
         },
         {
+            "name": "ConversionsParametersStudy",
+            "experiments": [
+                {
+                    "name": "NewDefaultPattern",
+                    "probability_weight": 100,
+                    "parameters": [
+                        {
+                            "name": "conversions_default_conversion_id_pattern",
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -830,6 +830,32 @@
                 "platform": ["WINDOWS", "MAC", "LINUX", "ANDROID"]
             }
         },
+        {
+            "name": "ConversionsParametersStudy",
```
tmancey, line 834: nit: `ConversionsStudy` as we normally just append `Study` to the feature name

tmancey: iOS does not support field trials at this time. We should fix this ASAP on brave-core .

iambrianfung: @tmancey i carried the name over from what @moritzhaller had originally on `main`.

kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on `main` (staging) as per https://github.com/brave/brave-variations/pull/108#issuecomment-921236785.

PR #93, opened 2021-08-09
Update Brave Talk study to be in Release (staging)

bsclifton:

PR #119, opened 2021-09-07
Update Brave Talk study to be in Release

bsclifton:
Per today's triage (Sep 7th), here's the uplift for Talk to RELEASE channel

cc: @jsecretan for the go-ahead. Was uplifted / tested against staging here:
https://github.com/brave/brave-variations/pull/93

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -913,7 +913,7 @@
                 }
             ],
             "filter": {
-                "channel": ["BETA", "NIGHTLY"],
+                "channel": ["RELEASE", "BETA", "NIGHTLY"],
                 "platform": ["WINDOWS", "MAC", "LINUX"]
             }
         }


rebron (approved):
Looks good. ++

seed/seed.json
```diff
@@ -913,7 +913,7 @@
                 }
             ],
             "filter": {
-                "channel": ["BETA", "NIGHTLY"],
```
iefremov, line 916: typically we should specify min_version, to not enable it for old clients @bsclifton @rebron

iefremov: also @kjozwiak

kjozwiak: @iefremov can we get this fixed or is it too late at this point? What would happen if a client that ended up pulling the above study doesn't meet the new requirements. Would it simply be removed? Assuming it wouldn't cause any issues for users that meet the requirements and have the study?

I'll make sure we have a `min_version` next time 👍

bsclifton: Thanks for the heads up - will do that going forward. No action needed for this, AFAIK. But they would be shown as being in an experiment which isn't applicable

PR #120, opened 2021-09-14
Rollout custom ad notifications

tmancey:
Rollout custom ad notifications on the release channel to:

- Windows 10 build 16299 and older or if native notifications are disabled
- macOS 10.13 and older
- Android 7 and older
- All versions of Linux

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -548,6 +548,10 @@
                     "probability_weight": 30,
                     "parameters": [
                         {
+                            "name": "can_fallback_to_custom_notifications",
+                            "value": "true"
+                        },
+                        {
                             "name": "ad_notification_normalized_display_coordinate_x",
                             "value": "1.0"
                         },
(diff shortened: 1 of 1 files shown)


SergeyZhukovsky (approved):
++

moritzhaller (approved):
LGTM as per screen share

seed/seed.json
```diff
@@ -589,7 +593,37 @@
             ],
             "filter": {
                 "min_version": "92.1.30.19",
-                "channel": ["NIGHTLY", "BETA"],
+                "min_os_version": "10.0.17134*",
```
goodov, line 596: looks like this is incorrect, see ```TEST(VersionTest, CompareToWildcardString)```.
https://github.com/chromium/chromium/blob/96.0.4643.1/base/version_unittest.cc#L148-L196

seed/seed.json
```diff
@@ -589,7 +593,37 @@
...
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 0
+                }
+            ],
+            "filter": {
+                "min_version": "92.1.30.19",
+                "max_os_version": "10.0.16299*",
```
goodov, line 625: same here

seed/seed.json
tmancey: Fixed

seed/seed.json
tmancey: Fixed

btlechowski:
Verification passed on

Brave | 1.30.84 Chromium: 94.0.4606.54 (Official Build) (64-bit)
-- | --
Revision | c8191a1d5cccbf64e8fe7269043f8ace8d74dd05-refs/branch-heads/4606@{#1130}
OS | Ubuntu 18.04 LTS

Verified Custom ads are enabled by default on Linux

image|image
-- | --

seed/seed.json
iefremov: @goodov unfortunately, our converter doesn't support this filter :-/ 
we definitely need better validation for `seed.json`

tmancey: Added support for min_os_version and max_os_version via https://github.com/brave/brave-variations/pull/126

PR #122, opened 2021-09-22
Enable NetworkIsolationKey-based partitions.

goodov:
https://github.com/brave/brave-browser/issues/11816

Enable the same features Chromium currently is rolling out. At first test it on nightly.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -1158,7 +1158,6 @@
                     "feature_association": {
                         "enable_feature": [
                             "PartitionConnectionsByNetworkIsolationKey",
-                            "PartitionDomainReliabilityByNetworkIsolationKey",
                             "PartitionExpectCTStateByNetworkIsolationKey",
                             "PartitionHttpServerPropertiesByNetworkIsolationKey",
                             "PartitionSSLSessionsByNetworkIsolationKey",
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -1074,6 +1074,34 @@
...
+        {
+            "name": "PartitionConnectionsByNetworkIsolationKeyStudy",
+            "experiments": [
+                {
+                    "name": "Enabled",
+                    "probability_weight": 100,
+                    "feature_association": {
+                        "enable_feature": [
+                            "PartitionConnectionsByNetworkIsolationKey",
+                            "PartitionDomainReliabilityByNetworkIsolationKey",
```
iefremov, line 1087: @pes10k can you please confirm we need `PartitionDomainReliabilityByNetworkIsolationKey`? just not sure we have it enabled

iefremov:
@goodov there is a conflict

seed/seed.json
pes10k: I think you're right, we shouldn't have the domain reliability functionality enabled (if we haven't already disabled it), and so we don't need the `PartitionDomainReliabilityByNetworkIsolationKey` feature either

pes10k: https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/domain_reliability/README.md if others aren't familiar with what the feature does (it reports to google when google properties are down)

goodov: agree. the domain reliability feature is disabled on our end.

iefremov (approved):

PR #131, opened 2021-09-27
Enable NetworkIsolationKey-based partitions

goodov:
https://github.com/brave/brave-browser/issues/11816

Enable the same features Chromium currently is rolling out. At first test it on nightly.

PR #136, opened 2021-10-04
Increase Split/Partition*ByNetworkIsolationKey to 100% on nightly.

goodov:
Increasing rollout to 100% on nightly as was discussed in Slack.

https://github.com/brave/brave-browser/issues/11816
See also: https://github.com/brave/brave-variations/pull/131

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -1048,7 +1048,7 @@
             "experiments": [
                 {
                     "name": "Enabled",
-                    "probability_weight": 50,
+                    "probability_weight": 100,
                     "feature_association": {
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. We'll leave `PartitionConnectionsByNetworkIsolationKeyStudy` enabled at 100% for a few days before we consider moving this onto `BETA`. We'll be doing a slower rollout on `BETA` as we have a bigger population compared to Nightly. So we'll start at ~25% and slowly work our way to 100%.

PR #143, opened 2021-10-15
Enable Ad Serving Version 2 for 50% of Nightly users

moritzhaller:


seed/seed.json
```diff
@@ -959,7 +959,67 @@
...
+                            "name": "ad_serving_version",
+                            "value": "2"
+                        }
+                    ],
+                    "feature_association": {
+                        "enable_feature": ["AdServing"]
+                    }
+                }
+            ],
+            "filter": {
```
moritzhaller, line 1021: @tmancey should we add `"min_version": "94.1.32.17"`? We might another experiment for lower Nightly versions then

kjozwiak:
Waiting till https://github.com/brave/brave-variations/pull/143#discussion_r729528116 gets addressed. Once this is merged into `main`, we'll get this QA'd and merged into `production`.

@tmancey in your opinion, do you think 50% in Nightly right away is aggressive? Should we start at ~25%? If something does go wrong/regresses, what are the side affects?

tmancey:
@kjozwiak this is a test we are only running on nightly at this time. If/when we believe this should go to production we will raise a new PR and reduce the rollout. Does this work for you?

PR #162, opened 2021-11-13
Enable "RequestAdsEnabledApi" feature on 100% in Release (prod).

aseren:


kjozwiak (approved):
Uplift into `production` at `100%` approved after deliberating with @brave/uplift-approvers.

PR #166, opened 2021-11-17
Ad notification timeout study

tmancey:
Field trial to lower the push notification ad timeout on desktop

-  Default (120 seconds, 25% of users)
-  60 seconds (25% of users)
-  30 seconds (25% of users)
-  15 seconds (25% of users)

PR #174, opened 2021-11-22
Ad notification timeout study

tmancey:


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on Nightly as per https://github.com/brave/brave-variations/pull/166#issuecomment-974504658.

PR #180, opened 2021-12-10
Increase partitioned blink memory cache to 100% on nightly.

goodov:
Increase to 100% on nightly.

https://github.com/brave/brave-browser/issues/18831


kjozwiak (approved):
Uplift into `production` at `100%` approved after deliberating with @brave/uplift-approvers.

PR #188, opened 2021-12-23
Enable BraveAdblockCookieListDefault on Nightly/Beta (staging)

antonok-edm:
This feature enables `Easylist-Cookie List - Filter Obtrusive Cookie Notices` from brave://adblock by default, although it can still be overridden by a user who wishes to disable the list.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -30,6 +30,34 @@
             }
         },
         {
+            "name": "CookieListDefaultStudy",
+            "experiments": [
+                {
+                    "name": "Enabled",
+                    "probability_weight": 100,
+                    "feature_association": {
+                        "enable_feature": ["BraveAdblockCookieListDefault"]
+                    }
+                },
+                {
+                    "name": "Disabled",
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -29,6 +29,34 @@
...
+                    "name": "Enabled",
+                    "probability_weight": 100,
+                    "feature_assocatiation": {
+                        "enable_feature": ["BraveAdblockCookieListDefault"]
+                    }
+                },
+                {
+                    "name": "Disabled",
+                    "probability_weight": 100,
+                    "feature_assocatiation": {
```
pes10k, line 45: this looks funny to me.  This should be 0 right?

pes10k (changes requested):

pes10k (approved):

seed/seed.json
antonok-edm: yes, also I noticed that somehow I managed to copy-paste `feature_assocatiation` :grimacing: 

should be fixed now

PR #191, opened 2021-12-27
Enable PartitionBlinkMemoryCacheStudy on 5% release.

goodov:
Release 5%.

https://github.com/brave/brave-browser/issues/18831

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -832,6 +832,29 @@
             }
         },
         {
+            "name": "PartitionBlinkMemoryCacheStudy",
+            "experiments": [
+                {
+                    "name": "Enabled",
+                    "probability_weight": 5,
+                    "feature_association": {
+                        "enable_feature": [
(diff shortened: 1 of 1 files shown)


kjozwiak:
@goodov @pes10k assuming we need to wait till https://github.com/brave/brave-browser/issues/18831 is on the `release` channel via `1.34.x` before we push this live on `production` via `Griffin`?

pes10k:
I don’t think we strictly need to, since if we push the giffin change before the code change makes it to stable, it’ll no op until the code catches up, but no strong preference on my end

kjozwiak:
As per the discussion on Slack with @pes10k & @goodov, we'll wait till `1.34.x` is live and then push this into `production` and enable ` PartitionBlinkMemoryCacheStudy` on `RELEASE` at `5%`.

kjozwiak (approved):
Uplift into `production` at `5%` approved after deliberating with @brave/uplift-approvers.

PR #205, opened 2022-01-14
Disable CNAME uncloacking on Android.

iefremov:


pes10k:
@iefremov can you say more about why this change is needed?

antonok-edm (approved):

kjozwiak:
> @iefremov can you say more about why this change is needed?

@pes10k see https://github.com/brave/brave-variations/pull/200#issue-1100440158. It has a link to the conversation re: the reason we're disabling it. Believe it's causing hangs/crashes on https://www.tripadvisor.com. @samartnik @bridiver @antonok-edm took a look and couldn't figure out was going on at a quick glance so this will need some investigation. Rather than leaving it and having users crash/hang, we'll disable it for now on Android until we can fix it.

Desktop is affected as well but it doesn't cause Brave to crash like Android does.

pes10k:
I see, understood @kjozwiak , thank you for the background (and thank you to @antonok-edm who explained similarly on Slack).  Sounds good, as long as it is a temporary measure, while the underlying bug is sorted out, no objections from me

pes10k (approved):

kjozwiak:
Verification **PASSED** on `Pixel 6` running `Android 12` using `1.34.80 Chromium: 97.0.4692.71`


Reproduced the original issue/crash by loading https://www.tripadvisor.com/Hotel_Review-g1798615-d252805-Reviews-Marriott_s_Timber_Lodge-South_Lake_Tahoe_Lake_Tahoe_California_California.html and refreshing the browser several times. Eventually Brave will crash as per 

Verified that the crash wasn't occurring once `DisableCnameUncloakingForAndroid:Disabled` as per the following:

Launched the daemon as per the following:

```
kjozwiak@ubuntu:~$ adb shell "echo 'https://variations.bravesoftware.com/seed' > /data/local/tmp/brave-test-variations-server-url"
* daemon not running; starting now at tcp:5037
* daemon started successfully
```

Confirmed that the daemon was running as per the following:

```
kjozwiak@ubuntu:~$ adb shell "cat /data/local/tmp/brave-test-variations-server-url"
https://variations.bravesoftware.com/seed
```

Ensured that `DisableCnameUncloakingForAndroid:Disabled` was being displayed under `brave://version` as per the following:

Screenshot_20220118-011834

Once the above was completed, refreshed https://www.tripadvisor.com/Hotel_Review-g1798615-d252805-Reviews-Marriott_s_Timber_Lodge-South_Lake_Tahoe_Lake_Tahoe_California_California.html ~30 different times and ensured Brave wasn't crashing.

kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on Nightly as per https://github.com/brave/brave-variations/pull/205#issuecomment-1015106670.

PR #207, opened 2022-01-17
Fix HangWatcher feature name

atuchin-m:
PR to main: https://github.com/brave/brave-variations/pull/206

The root issue:
brave/brave-browser#20304


kjozwiak (approved):

PR #214, opened 2022-02-02
Remove v2 Brave Ads serving pipeline from the beta channel

tmancey:
Remove v2 Brave Ads serving pipeline from the beta channel until this feature has been fully tested on nightly cc @kjozwiak @stephendonner

PR #215, opened 2022-02-02
Remove v2 Brave Ads serving pipeline from the beta channel

tmancey:
staging PR: https://github.com/brave/brave-variations/pull/214


tmancey:
Clsoing as not required as only affected staging

PR #216, opened 2022-02-10
Disable NavigationThreadingOptimizations on Desktop

petemill:
This is currently disabled on Chrome, at least for version 98 and for desktop. I am unsure if this problem affects Android or when we update to Chromium 99, so I've left it as default for android and desktop with cr99+ (enabled in brave-core/chromium).

Fix: https://github.com/brave/brave-browser/issues/21029


bsclifton (approved):
Nice work tracking this down! 😄

PR #217, opened 2022-02-11
Disable NavigationThreadingOptimizations on Desktop for 5% of users - Production

bsclifton:
## Test plan
To reproduce (on a bad profile). Run the following WITH and WITHOUT the study:
1. open [news.ycombinator.com](http://news.ycombinator.com/)
2. open devtools
3. go to network tab
4. check "disable cache"
5. Refresh
6. Observe gap between first request and subsequent requests

Here is a bad profile you can use 😄 
[slow-profile.zip](https://github.com/brave/brave-variations/files/8051775/slow-profile.zip)

## Description
Uplift of staging version merged with https://github.com/brave/brave-variations/pull/216 / https://github.com/brave/brave-variations/pull/218 / https://github.com/brave/brave-variations/pull/219

NOTE: do not merge yet. We need to finish testing and figure out what percentages to roll out

> This is currently disabled on Chrome, at least for version 98 and for desktop. I am unsure if this problem affects Android or when we update to Chromium 99, so I've left it as default for android and desktop with cr99+ (enabled in brave-core/chromium).
> 
> Fix: https://github.com/brave/brave-browser/issues/21029

PR #220, opened 2022-02-11
Remove `max_version` for `NavigationThreadingOptimizationsCompat` study (Staging) 

bsclifton:
This is being removed in https://github.com/brave/brave-variations/pull/217 (production)- want to match the same criteria in staging


bsclifton:
Merging so that staging matches production

PR #260, opened 2022-04-15
Disable DeAMP feature flag on production

ShivanKaul:
Disable DeAMP until blog post on Tuesday as per https://bravesoftware.slack.com/archives/G15HNQKV5/p1650006647890279?thread_ts=1649351034.570879&cid=G15HNQKV5

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -58,6 +58,34 @@
             }
         },
         {
+            "name": "DeAMPStudy",
+            "experiments": [
+                {
+                    "name": "Enabled",
+                    "probability_weight": 0,
+                    "feature_association": {
+                        "enable_feature": ["BraveDeAMP"]
+                    }
+                },
+                {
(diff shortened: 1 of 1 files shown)


bsclifton (approved):
Approving so that we'll be in a disabled state over the weekend 😄

seed/seed.json
```diff
@@ -57,6 +57,34 @@
...
+                        "disable_feature": ["BraveDeAMP"]
+                    }
+                },
+                {
+                    "name": "Default",
+                    "probability_weight": 0
+                }
+            ],
+            "filter": {
+                "min_version": "100.1.39.5",
```
iefremov, line 83: shouldn't it be 101. ?

ShivanKaul: Hmm, I thought I'd taken it from brave://system and 101 sounds right. Should I make the change @iefremov?

PR #264, opened 2022-05-13
Speedreader 100% in all channels.

iefremov:
There is no sense in keeping speedreader enabled for 40% in release channel. As the feature is more or less stable now, I think we should push to 100%


kjozwiak (approved):
Uplift into `Production` approved after deliberating with @brave/uplift-approvers.

PR #268, opened 2022-06-02
Re-enable cookie dialog blocking list on Nightly

antonok-edm:
Issues with a handful of websites were resolved with changes in `brave-core`. This PR re-enables the list by default on Nightly as per discussion on https://bravesoftware.slack.com/archives/C6YNM6Y5S/p1654120574558859?thread_ts=1654117214.714209&cid=C6YNM6Y5S

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -90,14 +90,14 @@
             "experiments": [
                 {
                     "name": "Enabled",
-                    "probability_weight": 0,
+                    "probability_weight": 100,
                     "feature_association": {
                         "enable_feature": ["BraveAdblockCookieListDefault"]
                     }
                 },
                 {
                     "name": "Disabled",
-                    "probability_weight": 100,
+                    "probability_weight": 0,
                     "feature_association": {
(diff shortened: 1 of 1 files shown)


antonok-edm:
Updated to just enable on Nightly, at least until https://github.com/brave/brave-core/pull/12827 makes it to Beta.

kjozwiak:
Verified as per https://github.com/brave/brave-variations/pull/268#issuecomment-1151342816 and ready to merge. However, the above is pending further discussion. Adding the `Uplift/Blocked` label.

kjozwiak:
@antonok-edm should we close this off for now till we have more info? Seems like we're going to need some client side changes re: having modals that let users enable this feature themselves rather than us doing it by default. I don't mind leaving this opened either. Let me know what works for you 👍

kjozwiak:
Going to close this off as we're going to go with the approach of adding a modal/some type of user interaction that will let them enable the list when Brave detects a cookie permission modal.

PR #277, opened 2022-06-09
EarlyBodyLoadCompat to nightly 100%

atuchin-m:
For https://github.com/brave/brave-browser/issues/23267

PR is created in advance to solve the issues to page loading. I suppose that disabling the feature should helps, but it's unsafe to switch it for the all the users. Therefore, the current PR disables the feature only for nightly users.

We still need more confirmations that it helps before merging.


atuchin-m:
Reverted from main in https://github.com/brave/brave-variations/pull/278

PR #284, opened 2022-06-14
Enabling NavigationThreadingOptimizations: BETA 50%, STABLE 5%

atuchin-m:
For https://github.com/brave/brave-browser/issues/23267

Now we have 3 non-overlapping groups:
Two `NavigationThreadingOptimizationsCompat`: one for BETA (50%/50%) and one of new STABLE >=cr102 (5%/95%). We will remove all these group in the next few days to make the feature be in the default (enabled) state.

The last one is`NavigationThreadingOptimizationsCompatOldVersions`: a group for old clients (<cr102) to continue disabling the feature for them (because of a bug with navigations/localStorage). This group will stay in the config for some long period of time (~one year).

PR #287, opened 2022-06-15
NavigationThreadingOptimizations BETA 100%, STABLE 25%

atuchin-m:
For https://github.com/brave/brave-browser/issues/23267

BETA section is removed.
STABLE section is changed: 5/95 => 25/75

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -1071,7 +1071,7 @@
             "experiments": [
                 {
                     "name": "Disabled",
-                    "probability_weight": 50,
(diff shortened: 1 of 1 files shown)


iefremov (approved):

kjozwiak (approved):
Uplift into `production` approved 👍 QA verified the above study via https://github.com/brave/brave-variations/pull/284#issuecomment-1156031357 so I don't think there's a need to check again as we're just increasing the % on `BETA`  & `RELEASE`.

PR #295, opened 2022-06-23
Revert "NTP SI User Activity"

tmancey:
Reverts brave/brave-variations#293


kjozwiak (approved):
Uplift into `main` approved 👍

PR #296, opened 2022-06-24
Revert "User activity ntp si prod"

tmancey:
Reverts brave/brave-variations#294


kjozwiak (approved):
Uplift into `production` approved 👍

PR #301, opened 2022-07-07
Increase reduce language to 100/100/30

pilgrim-brave:


kjozwiak (approved):

PR #302, opened 2022-07-07
Revert "Enables Vulkan on Android for nightly channel with a black list (PRODUCTION)"

kjozwiak:
Reverts brave/brave-variations#300


SergeyZhukovsky (approved):
++

PR #315, opened 2022-07-14
make reduce lang 100% on all platforms

pes10k:


kjozwiak (approved):

PR #320, opened 2022-07-19
releasing ads serving v2 to 10% of release

ptjames:


kjozwiak (approved):

Issue #338, opened 2022-08-25
Enable translate in Beta using Griffin


atuchin-m:
Nightly issue: https://github.com/brave/brave-browser/issues/24957
Slack thread: https://bravesoftware.slack.com/archives/C01LKMP6X36/p1661365425546619

bsclifton:
Closing as this should be good to go 😄

Issue #335, opened 2022-08-24
Enable BraveTranslateStudy to Release 25%


rebron:
## Description 

* Enable `BraveTranslateStudy` to `Release` at `25%`
* Enabled Features: `UseBraveTranslateGo`

Rollout will be `25%` to `50%` to `75%` to `100%` and we will wait two days in between next % rollout.

Channels: `Release`
Countries: `All`
Platforms: `Windows`, `Mac`, `Linux` & `Android`

## Other Additional Information:

`BraveTranslateStudy` should be enabled to countries: all; channels: `Nightly` and `Beta` at 100%; and platforms: `windows`, `mac`, `linux` & `Android`

kjozwiak:
Verified the `25%` phase of the rollout on `RELEASE` via https://github.com/brave/brave-variations/pull/343#issuecomment-1234350258.

bsclifton:
Closing as this is released to 25% 👍

PR #339, opened 2022-08-25
Android translate to BETA 100%

atuchin-m:
Related https://github.com/brave/brave-variations/issues/338

PR #340, opened 2022-08-25
Android translate to BETA 100%

atuchin-m:
Resolves https://github.com/brave/brave-variations/issues/338

The PR enables translate in Android beta and merge desktop and Android studies into the one. 

The following things have done to support merging:
1. Min version was bumped to "104.1.43.54" (we need https://github.com/brave/brave-browser/issues/23770) It covers beta/nightly builds not older than 3 weeks ago. IMHO it's okay: we shouldn't about older non-stable installations. 
[The proof that 1.43.54 has the commit](https://user-images.githubusercontent.com/45488748/186644504-b25da4b2-cc8a-4e16-ac6c-b41ad6044aff.png)

2. Group "Disabled" was removed (because "probability_weight": 0 and we can't explicitly disable `Translate` feature for desktops);

3. Explicitly enabling `Translate` feature is added. It is enabled by default for Desktop([link](https://github.com/chromium/chromium/blob/main/components/translate/core/browser/translate_prefs.cc#L180)), but not for Android ([link](https://github.com/brave/brave-core/blob/master/chromium_src/components/translate/core/browser/translate_prefs.cc#L22)).

The next step is enable both desktop/android in release: https://github.com/brave/brave-variations/issues/335

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -1247,14 +1247,7 @@
                     "name": "Enabled",
                     "probability_weight": 100,
                     "feature_association": {
-                        "enable_feature": ["UseBraveTranslateGo"]
-                    }
-                },
-                {
-                    "name": "Disabled",
-                    "probability_weight": 0,
-                    "feature_association": {
-                        "disable_feature": ["UseBraveTranslateGo"]
+                        "enable_feature": ["UseBraveTranslateGo", "Translate"]
                     }
                 },
                 {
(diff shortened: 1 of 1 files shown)


iefremov (approved):

kjozwiak (approved):
Uplift into `production` approved 👍 QA has verified the PR on Nightly as per https://github.com/brave/brave-variations/pull/339#issuecomment-1231119539.

PR #342, opened 2022-08-30
Limit studies that already enabled in code to current beta version.

goodov:
These features are all enabled in code, so limiting these studies to the current "beta" to remove it from the seed when `max_version` will be phased out.
https://github.com/brave/brave-browser/issues/24148

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -311,6 +311,7 @@
             ],
             "filter": {
                 "min_version": "90.1.25.57",
+                "max_version": "104.1.43.79",
                 "channel": ["NIGHTLY", "BETA"],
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -363,26 +365,7 @@
           ],
           "filter": {
               "min_version": "101.1.38.109",
-              "channel": ["NIGHTLY", "BETA", "RELEASE"],
-              "platform": ["WINDOWS", "MAC", "LINUX", "ANDROID"]
-          }
-        },
-        {
-          "name": "DisableBlobPartitioning",
```
goodov, line 371: this study is removed, because `102.1.39.117` is already phased out.

PR #358, opened 2022-09-08
Enables Vulkan for 10% users on a stable channel

SergeyZhukovsky:
Resolves https://github.com/brave/brave-browser/issues/25299

PR #359, opened 2022-09-08
Enables Vulkan for 10% users on a stable channel

SergeyZhukovsky:
Resolves https://github.com/brave/brave-browser/issues/25299


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on `main` (staging) as per https://github.com/brave/brave-variations/pull/358#issuecomment-1244135317.

PR #370, opened 2022-09-17
Enables Vulkan for 50% users on a stable channel

SergeyZhukovsky:


kjozwiak (approved):

PR #399, opened 2022-10-17
Enables bg playback on Android to 100% percent of users on nightly channel

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/398


kjozwiak (approved):
LGTM 👍 Approving/Merging into `main` (staging) so we can quickly run through it and uplift https://github.com/brave/brave-variations/pull/400 into `production`.

PR #400, opened 2022-10-17
Enables bg playback on Android to 100% percent of users on nightly channel(production)

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/398


kjozwiak (approved):

PR #401, opened 2022-10-20
Rollout v2 ad serving pipeline to 50% of users

tmancey:
This feature has been baked into the release channel at 10% for some time. We now need to increase the rollout to 50% of users.


seed/seed.json
```diff
@@ -390,7 +390,7 @@
             "experiments": [
                 {
                     "name": "DefaultAdNotificationsPerHour=10/MaximumAdNotificationsPerDay=100/MaximumInlineContentAdsPerHour=12/MaximumInlineContentAdsPerDay=60/AdServingVersion=1",
-                    "probability_weight": 90,
+                    "probability_weight": 50,
```
goodov, line 393: you need to change other experiment in this study so overall weight would become 100.

tmancey: Great catch, thanks. Fixed

goodov (approved):

PR #408, opened 2022-10-25
enable BraveAdblockCookieListOptIn on release channel at 10% [`production`]

antonok-edm:
`main`: https://github.com/brave/brave-variations/pull/407

PR #414, opened 2022-10-27
enable BraveAdblockCookieListOptIn on release channel at 25% [`production`]

antonok-edm:
followup to https://github.com/brave/brave-variations/pull/408


kjozwiak (approved):

PR #420, opened 2022-11-03
Enables background playback to all users on beta channel on Android

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/419


kjozwiak (approved):
LGTM 👍 Approving so I can run through the above on `staging` and get https://github.com/brave/brave-variations/pull/421 verified via `production`.

PR #422, opened 2022-11-03
enable BraveAdblockCookieListOptIn on release channel at 100% [`main`]

antonok-edm:
`production`: https://github.com/brave/brave-variations/pull/423

PR #423, opened 2022-11-03
enable BraveAdblockCookieListOptIn on release channel at 100% [`production`]

antonok-edm:
`main`: https://github.com/brave/brave-variations/pull/422


kjozwiak (approved):

PR #439, opened 2022-11-07
Auto upsert-study

kim0:
Resolves https://github.com/brave/devops/issues/8609

### DIFF
--- a/.github/ci/griffin-study-utils.py
+++ b/.github/ci/griffin-study-utils.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+import json
+import sys
+
+
+def _load_studies(filename="seed/seed.json"):
+    with open(filename) as fh:
+        studies = json.load(fh)
+    return studies
+
+
+def _save_studies(studies, filename="seed/seed.json"):
+    with open(filename, "w") as fh:
+        json.dump(studies, fh, sort_keys=True, indent=4)
+
+
+def _create_study(
+    name: str,
+    enable_feature: str,
+    probability_enabled: int,
+    channel: list[str],
+    platform: list[str],
+    min_version: str,
+) -> dict:
+
+    study = {
+        "name": name,
+        "experiments": [
+            {
+                "name": "Enabled",
+                "probability_weight": probability_enabled,
+                "feature_association": {"enable_feature": [enable_feature]},
+            },
+            {"name": "Default", "probability_weight": 100 - probability_enabled},
+        ],
+        "filter": {
+            "channel": channel,
+            "platform": platform,
+        },
+    }
+    if min_version:
+        study["filter"]["min_version"] = min_version
+    return study
+
+
+def _upsert_study(
+    name: str,
+    enable_feature: str,
+    probability_enabled: int,
+    channel: list[str],
+    platform: list[str],
+    min_version: str,
+):
+    rawstudies = _load_studies()
+    studies: list[dict] = rawstudies["studies"]
+    study_to_add = _create_study(
+        name,
+        enable_feature,
+        probability_enabled,
+        channel,
+        platform,
+        min_version,
+    )
+    existing_study = False
+    for idx, study in enumerate(studies):
+        if study["name"] == name:
+            existing_study = True
+            studies[idx] = study_to_add
+            # print(studies[idx])
+    if not existing_study:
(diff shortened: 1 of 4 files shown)


.github/ci/griffin-study-utils.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+import json
```
mihaiplesa, line 2: Is there usually a newline between line 1 and 2 here?

.github/workflows/upsert-study.yml
```diff
@@ -0,0 +1,85 @@
...
+        required: true
+        default: "RELEASE,BETA,NIGHTLY"
+        type: string
+      platform:
+        description: "Platforms to enable on, in comma separated value format"
+        required: true
+        default: "WINDOWS,MAC,LINUX,ANDROID"
+        type: string
+      min_version:
+        description: "Minimum version to enable on [Optional]"
```
mihaiplesa, line 28: `(optional)`

.github/workflows/upsert-study.yml
```diff
@@ -0,0 +1,85 @@
...
+        type: string
+      channel:
+        description: "Channels to enable on, in comma separated value format"
+        required: true
+        default: "RELEASE,BETA,NIGHTLY"
+        type: string
+      platform:
+        description: "Platforms to enable on, in comma separated value format"
+        required: true
+        default: "WINDOWS,MAC,LINUX,ANDROID"
```
mihaiplesa, line 25: Sort?

.github/workflows/upsert-study.yml
```diff
@@ -0,0 +1,85 @@
...
+        description: "Feature name to enable"
+        required: true
+        type: string
+      probability_enabled:
+        description: "Percentage number to enable. ex: 25"
+        required: true
+        default: "25"
+        type: string
+      channel:
+        description: "Channels to enable on, in comma separated value format"
```
mihaiplesa, line 18: `(in comma separated value format)`

.github/workflows/upsert-study.yml
```diff
@@ -0,0 +1,85 @@
...
+      name:
+        description: "Study name"
+        required: true
+        type: string
+      enable_feature:
+        description: "Feature name to enable"
+        required: true
+        type: string
+      probability_enabled:
+        description: "Percentage number to enable. ex: 25"
```
mihaiplesa, line 13: `(ex: 25)`

mihaiplesa (changes requested):

.github/ci/griffin-study-utils.py
kim0: `autopep8` is not adding any, and linter is happy. So I don't think so

.github/workflows/upsert-study.yml
kim0: Done

.github/workflows/upsert-study.yml
kim0: Don't think we should here. The json file contains this particular order everywhere, and I think it's better to add the new ones in the same order

mihaiplesa (approved):

kim0:
@antonok-edm can you please review this PR. This was done as a Github Action, since we're generally moving away from Jenkins. Also I had to reformat the `seed.json` file since it gets parsed and rewritten every time. So to get good diffs, the order and formatting needs to be deterministic. I hope this will be OK.

I have used this command to auto-format `python3 -m json.tool --sort-keys --indent 4 seed/seed.json seed/seed.json.out && mv seed/seed.json.out seed/seed.json` bec it is compatible with the output from the `.github/ci/griffin-study-utils.py` script.

Here are two sample output PRs #448 and #447 .. They contain unneeded files, just bec this PR is not yet merged. Once merged, they should only contain changes from `seed/seed.json`

Also, here's how the slack notification message looks like
slack-notification-griffin

Since GHA does not allow multi-select choice boxes, I ended up using comma separated values like `WINDOWS,MAC,LINUX`. I think we have to get this PR merged before we can see the input UI. Let me know if any changes is needed!

antonok-edm (approved):
This is really neat, great work @kim0 !

mihaiplesa (approved):

Issue #411, opened 2022-10-25
Roll out BraveBlockScreenFingerprinting to Desktop Release users


arthuredelstein:
Now we have rolled out to 100% of Nightly and Beta Desktop users, we are ready to roll out to Release. The plan is to do the following steps:
* [x] 10%
* [x] 25%
* [x] 50%
* [ ] 100%

After we get to 100%, and assuming there are no problems, we will enable the flag by default and we can remove from brave-variations.

kjozwiak:
`BraveBlockScreenFingerprinting` was pushed to `production` at `10%` via https://github.com/brave/brave-variations/pull/413 on `Oct 27, 2022 @ 3:45am EST`.

PR #468, opened 2022-11-22
Complete rollout of BraveBlockScreenFingerprinting to 100% of Release Desktop users

arthuredelstein:
This is step 4 of https://github.com/brave/brave-variations/issues/411.

Notice it's showing a newline added at the end of the file! I tried creating this commit three different ways, including editing directly in github, and this newline change always appears. I don't understand the reason.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -1647,11 +1647,11 @@
                         ]
                     },
                     "name": "Enabled",
-                    "probability_weight": 50
+                    "probability_weight": 100
                 },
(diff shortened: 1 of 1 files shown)


kjozwiak:
Closing as we're going to roll this study/feature back due to some webcompat issues that were discovered. For example, https://www.hyatt.com not loading when `BraveScreenFingerprintingBlockerStudy:Enabled` via `brave://version`.

PR #469, opened 2022-11-22
Complete rollout of BraveBlockScreenFingerprinting to 100% of Release Desktop users

arthuredelstein:


kjozwiak (commented):
Closing as we're going to roll this study/feature back via https://github.com/brave/brave-variations/pull/473 due to some webcompat issues that were discovered. For example, https://www.hyatt.com not loading when `BraveScreenFingerprintingBlockerStudy:Enabled` via `brave://version`.

PR #485, opened 2022-12-16
Update inline content ads

tmancey:
Decrease inline content ads per hour from 12 to 6 and ads per day from 60 to 20

PR #488, opened 2022-12-19
Update inline content ad names to reflect inline content ads per hour/day

tmancey:
Followup to https://github.com/brave/brave-variations/pull/485

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -526,7 +526,7 @@
                             "AdServing"
                         ]
                     },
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Thanks for the fixing the above 👍 Approving & Merging into `main` so we can get this verified on `staging and get it uplifted into `production ` via https://github.com/brave/brave-variations/pull/489.

PR #489, opened 2022-12-19
Update inline content ads

tmancey:
See https://github.com/brave/brave-variations/pull/485 and https://github.com/brave/brave-variations/pull/488


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on staging as per https://github.com/brave/brave-variations/pull/485#issuecomment-1359635060.

Issue #519, opened 2023-02-07
Enable Brave news V2 on nightly for android


deeppandya:
Enables feature introduced in https://github.com/brave/brave-core/pull/16473 on production

deeppandya:
Closing this issue as irrelavent.

PR #520, opened 2023-02-07
Enable news v2 on nightly

deeppandya:
Resolves https://github.com/brave/brave-variations/issues/519


kjozwiak:
Going to close this off as discussed via the `Android` triage/planning meeting. We'll enable this via client side rather than Griffin as the feature is basically ready and should be enabled. CCing @mattmcalister.

PR #528, opened 2023-02-28
Disable ads for pre Rewards 2.5 builds

tmancey:
Ads should not be served on 109.1.48.60 or older
Ads should be served on 109.1.48.61 or newer

PR #529, opened 2023-02-28
Disable ads for pre Rewards 2.5 builds

tmancey:
See https://github.com/brave/brave-variations/pull/528


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on `staging` as per https://github.com/brave/brave-variations/pull/528#issuecomment-1449156357.

PR #539, opened 2023-03-10
Roll out Brave HTTPS by Default to Beta Desktop

arthuredelstein:


kjozwiak (approved):
Uplifting into `main` so the above can quickly get verified via staging so we can get https://github.com/brave/brave-variations/pull/540 uplifted into `production`.

PR #551, opened 2023-03-15
Enables AllowCertainClientHintsStudy to 100% on Release channel.

mkarolin:
Fixes https://github.com/brave/brave-variations/issues/550


kjozwiak (approved):
Approving & Merging so the above can checked via `main` (staging) before uplifting https://github.com/brave/brave-variations/pull/552 into `production`.

PR #556, opened 2023-03-16
remove de-amp, already enabled on all platforms

ShivanKaul:
https://github.com/brave/brave-core/blob/master/components/de_amp/common/features.cc#L16


kjozwiak (approved):

PR #557, opened 2023-03-17
Disable BraveCleanupSessionCookiesOnSessionRestore on MacOS.

goodov:
Brave restart after an update on MacOS doesn't work as expected. Because of this we cleanup session cookies when we shouldn't.

This PR restores the old behavior on MacOS until we fix the restart process.

https://github.com/brave/brave-browser/issues/29135

PR #558, opened 2023-03-17
Disable BraveCleanupSessionCookiesOnSessionRestore on MacOS.

goodov:
Brave restart after an update on MacOS doesn't work as expected. Because of this we cleanup session cookies when we shouldn't.

This PR restores the old behavior on MacOS until we fix the restart process.

https://github.com/brave/brave-browser/issues/29135

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2056,6 +2056,31 @@
                 ]
             },
             "name": "BraveAdblockMobileNotificationsListDefault"
+        },
+        {
+            "experiments": [
+                {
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on Nightly as per https://github.com/brave/brave-browser/issues/29135#issuecomment-1474056289 & https://github.com/brave/brave-variations/pull/557#issuecomment-1486160746.

PR #568, opened 2023-03-24
Enable mobile notifications list by default on main for release channel

antonok-edm:
in preparation for #508. #507 was missing release channel.


kjozwiak (approved):
LGTM 👍 Thanks @antonok-edm. Merging so can continue with https://github.com/brave/brave-variations/pull/507.

PR #575, opened 2023-03-30
Enable Https by Default in Android Beta

arthuredelstein:


kjozwiak (approved):
Merging into `main` so QA can verify that `HTTPS By Default` is being enabled on `BETA` before uplifting https://github.com/brave/brave-variations/pull/576 into `production`.

PR #596, opened 2023-04-13
Set NTP-SI ads per day to zero for May 5th to monitor metrics for non-opted-in users to use Eastern Time - Production

tmancey:
Uplift of https://github.com/brave/brave-variations/pull/595

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -628,8 +628,8 @@
             "filter": {
                 "channel": [
                     "RELEASE",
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on `main` (staging) as per https://github.com/brave/brave-variations/pull/603#issuecomment-1514723578.

PR #608, opened 2023-04-20
Add support for all Finch filters - Production

tmancey:
Uplift request for https://github.com/brave/brave-variations/pull/607

See https://github.com/brave/brave-variations/blob/main/seed/proto/study.proto for usage of each filter


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. 👍

PR #630, opened 2023-05-09
Bump up Google Sign In to 50% on Release

ShivanKaul:


kjozwiak (approved):
LGTM 👍 Merging into `main` so we can get https://github.com/brave/brave-variations/pull/631 merged into `production`. There isn't really anything QA needs to do in terms of verifications as we're simply increasing the percentage.

Issue #634, opened 2023-05-25
Set bg video playback in Enabled to 50% on Android


SergeyZhukovsky:
We don't see any issues with it enabled for 20% of users on a stable channel. Let's bump it up to 50%.

PR #636, opened 2023-05-25
Bumps bg video playback feature to 50 percent on Android

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/634


kjozwiak (approved):

PR #639, opened 2023-05-30
Roll out HTTPS by Default to 40% of Release on Desktop and Android

arthuredelstein:


kjozwiak (approved):

PR #643, opened 2023-05-31
fmt and check during CI

kim0:
Resolves https://github.com/brave/devops/issues/9809

### DIFF
--- a/.github/ci/griffin-study-utils.py
+++ b/.github/ci/griffin-study-utils.py
@@ -22,7 +22,6 @@ def _create_study(
     platform: list[str],
     min_version: str,
 ) -> dict:
-
     study = {
         "name": name,
         "experiments": [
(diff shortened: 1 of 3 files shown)


kim0:
cc @antonok-edm 

This adds the ability to `fmt` `seed.json` file. And uses that in the test pipeline. It runs `fmt` then asks git if there's any changes. If it finds changes, it shows the diff (telling the dev what needs to change) and fails [sample](https://github.com/brave/brave-variations/actions/runs/5131658311/jobs/9232017178?pr=643). If no changes are found then [success](https://github.com/brave/brave-variations/actions/runs/5131672686/jobs/9232045819?pr=643).

antonok-edm (approved):
Might be nice to have an easy-to-run script to format the file; right now this would have to be done using `python3 ./.github/ci/griffin-study-utils.py fmt` which is not the easiest to discover. But otherwise LGTM.

kim0:
Thanks @antonok-edm, I went ahead and merged it. The `fmt` command you mentioned can be wrapped in a one-line shell script if you prefer. I'll consider this done from my side, let me know if I can help with anything else.

PR #701, opened 2023-07-26
Disable SERP attribution on Android due to a crash - prod

tmancey:
Uplift of https://github.com/brave/brave-variations/pull/700


kjozwiak (approved):

PR #710, opened 2023-08-04
Increase rollout BraveP3AConstellation to 67% of release users

DJAndries:
https://github.com/brave/brave-variations/issues/709


kjozwiak (approved):
Approving & Merging and then we'll get https://github.com/brave/brave-variations/pull/711 uplifted into `production` on Monday.

PR #720, opened 2023-08-14
change countries and times

iambrianfung:
updating countries and times to match campaign metadata
cc: @ptjames @tmancey


tmancey (approved):
LGTM++

PR #736, opened 2023-09-08
Increase ModuleFileNamePatch to 100% on beta&nightly. (#735)

boocmp:


kjozwiak (approved):

PR #744, opened 2023-09-14
npm run serve:dev allows hot-reloading dev experience

petemill:
Starts web server, frontend gets changes as soon as files are saved. Uses webpack for typescript compilation. Puts all html and bundle output at src/web/public/ path


src/package.json
```diff
@@ -29,9 +29,11 @@
     "protobufjs": "^7.2.4",
     "protobufjs-cli": "^1.1.2",
     "style-loader": "^3.3.3",
+    "ts-loader": "^9.4.4",
```
atuchin-m, line 32: thanks.
I also used ts-loader, but decided to cut it, because we already use `tsc` for tracker.

src/web/webpack.config.js
```diff
@@ -34,6 +46,18 @@ module.exports = {
           name: '[name].[ext]',
         },
       },
+      {
+        test: /\.tsx?$/,
+        use: [
+          {
+            loader: require.resolve('ts-loader'),
```
atuchin-m, line 53: What `tsconfig.json` we use here?
If `src/tsconfig.json` then we could cut `src/bundle/tsconfig.json`.

atuchin-m (approved):

PR #749, opened 2023-09-25
Rollout TabSwitch by Default to 100% in Nightly on Desktop

fallaciousreasoning:

PR #750, opened 2023-09-25
Rollout TabSwitch by Default to 100% in Nightly on Desktop

fallaciousreasoning:
Staging PR https://github.com/brave/brave-variations/pull/749

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -74,6 +74,34 @@
                 {
                     "feature_association": {
                         "enable_feature": [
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the change on `Nightly` using `Win` on `staging` as per https://github.com/brave/brave-variations/pull/749#issuecomment-1745840331.

PR #800, opened 2023-11-03
Increase AI Chat enabled to 50% (uplift #799)

petemill:
Uplift https://github.com/brave/brave-variations/pull/799 from staging


kjozwiak (approved):

PR #814, opened 2023-11-08
Update for 1.60 which will show the Whats New page for folks

bsclifton:
Will show for anyone who was on version 1.59 or earlier

Will not show for people whose first version of Brave was 1.60.110 (since 1.60.110 and 1.60.114 are the same MAJOR version)

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2048,7 +2048,7 @@
                     "parameters": [
                         {
                             "name": "target_major_version_stable",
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved 👍 QA has verified the PR on Nightly as per https://github.com/brave/brave-variations/pull/812#issuecomment-1802519400.

PR #834, opened 2023-11-21
max_version for BraveAIChatEnabledStudy

petemill:
`BraveAIChatEnabledStudy` is at 100% and is enabled by default in brave-core via https://github.com/brave/brave-core/pull/20982, so set a `max_version` where that has landed


kjozwiak (approved):

PR #852, opened 2023-12-19
TTS enabled in release [25%].

boocmp:
https://github.com/brave/brave-variations/issues/851


kjozwiak (approved):

PR #853, opened 2023-12-19
TTS enabled in release [25%].

boocmp:
Resolves https://github.com/brave/brave-variations/issues/851


kjozwiak (approved):
LGTM 👍 Simple increase so there's not much that needs to be checked in terms of QA verification.

PR #868, opened 2024-01-15
[Brave News]: Enable FeedV2 in Release 1.62

fallaciousreasoning:
@kjozwiak I'd like to enable this just in 1.62.x but they're both on the same Chromium version, as best I can tell

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -75,6 +75,35 @@
                 {
                     "feature_association": {
                         "enable_feature": [
+                            "BraveNewsFeedUpdate"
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
LGTM 👍 Going to run through the above via `staging` and then we'll get a PR going for `production` and set it at `10%` which will be merged once `1.62.x` goes live.

fallaciousreasoning:
PR for enabling in 10% of Release:
https://github.com/brave/brave-variations/pull/881

Issue #874, opened 2024-01-17
Enable playlist android on nightly [Prod]


deeppandya:
Enable Playlist and PlaylistFakeUA features for Android on Nightly.

PR #873, opened 2024-01-16
Add playlist android study for Beta/Nightly on Staging

deeppandya:
https://github.com/brave/brave-variations/issues/872

PR #875, opened 2024-01-17
Add playlist android study for Nightly on Prod

deeppandya:
https://github.com/brave/brave-variations/issues/874

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2757,8 +2757,7 @@
             ],
             "filter": {
                 "channel": [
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the above PR(s) on `staging` using `Nighty` via `1.64.x` & `BETA` via `1.63.x` as per https://github.com/brave/brave-variations/pull/873#issuecomment-1905236855.

PR #884, opened 2024-01-22
Fix CI check `fmt` for trailing \n in seed.json

mherrmann:
The previous implementation flagged a trailing newline in seed.json as an error. Now, the file is normalized and expected to end in this way. This makes sure that editors that automatically append a newline do not cause accidental CI failures.

Resolves https://github.com/brave/devops/issues/11128.


mihaiplesa (approved):

kjozwiak:
@mherrmann @mihaiplesa do we need to get this into `production` as well? Or did this just affect `main`?

PR #885, opened 2024-01-23
Fix fmt check for trailing newline (uplift into `production`)

mherrmann:
Uplift of #884 into `production`.

The previous implementation flagged a trailing newline in seed.json as an error. Now, the file is normalized and expected to end in this way. This makes sure that editors that automatically append a newline do not cause accidental CI failures.


kjozwiak (approved):

PR #894, opened 2024-01-25
Rename Brave Ads experiments

aseren:
It is expected that experiment name doesn't contain a separator `/`: https://github.com/brave/chromium/blob/42d5c83e1ac9e48fc6eb44e1628ffe19a6ab26fe/base/metrics/field_trial.cc#L168

This expectation was introduced here: https://chromium-review.googlesource.com/c/chromium/src/+/4620752

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -561,7 +561,7 @@
                             "AdServing"
                         ]
                     },
-                    "name": "DefaultAdNotificationsPerHour=10/MaximumAdNotificationsPerDay=100/MaximumInlineContentAdsPerHour=6/MaximumInlineContentAdsPerDay=20/AdServingVersion=2",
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -2438,7 +2438,7 @@
                             "CustomNotificationAds"
                         ]
                     },
-                    "name": "normalized_display_coordinate_x/1/normalized_display_coordinate_y/0/inset_x/0/inset_y/72",
+                    "name": "normalized_display_coordinate_x|1|normalized_display_coordinate_y|0|inset_x|0|inset_y|72",
```
tmancey, line 2441: `name=value` instead of `name|value`, i.e., `foo=1|bar=2'

aseren: Fixed

tmancey (approved):
LGTM

kjozwiak:
@aseren @tmancey do these changes need to be applied into https://github.com/brave/brave-variations/blob/production/seed/seed.json as well? Looks like `production` is using `/` as well.

tmancey:
Yes please. @aseren can you please create an uplift request. Thanks

PR #896, opened 2024-01-26
Rename Brave Ads experiments

aseren:
Uplift request of https://github.com/brave/brave-variations/pull/894

Renamed experiments basing on upstream expectation that experiment name doesn't contain a separator `/`: https://github.com/brave/chromium/blob/42d5c83e1ac9e48fc6eb44e1628ffe19a6ab26fe/base/metrics/field_trial.cc#L168

This expectation was introduced here: https://chromium-review.googlesource.com/c/chromium/src/+/4620752


kjozwiak (approved):

PR #898, opened 2024-01-26
Cleanup CustomNotificationAds study parameters

aseren:
Custom notification ad griffin parameters are not needed, because they became obsolete and default ones are already used. 
Custom notification ad study is running for Nightly and Beta, so this change won't affect Release. These parameters were used to change position of first notification ad show only so it won't break notification ad for browsers with an old version.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2396,26 +2396,12 @@
                             "CustomNotificationAds"
                         ]
                     },
-                    "name": "normalized_display_coordinate_x=1|normalized_display_coordinate_y=0|inset_x=0|inset_y=220",
-                    "parameters": [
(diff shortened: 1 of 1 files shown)


tmancey (approved):
LGTM++

btlechowski:
Verified with

Brave | 1.63.147 Chromium: 121.0.6167.139 (Official Build) beta (64-bit)
-- | --
Revision | 714ec16e326c20dcdd5c70524509c652734689e8
OS | Windows 10 Version 22H2 (Build 19045.3930)


Verified CustomNotificationAdsStudy is enabled on beta
Verified the custom ad is positioned as expected 
image

PR #900, opened 2024-01-26
Cleanup CustomNotificationAds study parameters

aseren:
Uplift request for https://github.com/brave/brave-variations/pull/898

Custom notification ad griffin parameters are not needed, because they became obsolete and default ones are already used.
Custom notification ad study is running for Nightly and Beta, so this change won't affect Release. These parameters were used to change position of first notification ad show only so it won't break notification ad for browsers with an old version.


tmancey (approved):
LGTM

kjozwiak (approved):
Uplift into `production` approved 👍 QA has also quickly verified/sanity checked the changes via https://github.com/brave/brave-variations/pull/898#issuecomment-1932846924.

PR #911, opened 2024-01-29
Roll NewTabPageTakeoverPositionStudy for all platforms

aseren:


kjozwiak (approved):
LGTM 👍 Adding https://github.com/brave/brave-core/pull/21751 can be checked on `Desktop` & `Android`. Currently, `NewTabPageTakeoverPositionStudy` only has `IOS` as an available platform.

PR #936, opened 2024-02-23
Remove old/deprecated studies

goodov:


kjozwiak (approved):

Issue #941, opened 2024-02-28
Disable HeapProfilerReporting feature


atuchin-m:
Slack thread: https://bravesoftware.slack.com/archives/C05S50MFHPE/p1708949338401719

kjozwiak:
Closing as https://github.com/brave/brave-variations/pull/943#pullrequestreview-1909791062 was merged into `production`.

PR #942, opened 2024-02-28
Add HeapProfilingMacKillSwitch

atuchin-m:
For https://github.com/brave/brave-variations/issues/941

PR #943, opened 2024-02-28
Add HeapProfilingMacKillSwitch

atuchin-m:
Resolves https://github.com/brave/brave-variations/issues/941
PR to main: https://github.com/brave/brave-variations/pull/942

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2131,6 +2131,32 @@
                 ]
             },
             "name": "BraveRewardsAllowSelfCustodyProvidersStudy"
+        },
+        {
+            "experiments": [
+                {
+                    "feature_association": {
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA basically verified that `HeapProfilingMacKillSwitch` is only being enabled on `macOS` when both the `"min_version"` & `"max_version"` conditions are met as per https://github.com/brave/brave-variations/pull/942#issuecomment-1971662559. However, because the issue is locked/behind permissions, there's not much else QA can do in terms of verification. Some more context via https://bravesoftware.slack.com/archives/C01LKMP6X36/p1709108070563199.

PR #957, opened 2024-03-06
Enable HistoryMoreSearchResults in Nightly

fallaciousreasoning:
This flag lets more history search results show up in the omnibox

PR #958, opened 2024-03-06
Enable HistoryMoreSearchResults in Nightly

fallaciousreasoning:
This flag lets more history search results show up in the omnibox


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the above Griffin study on `main` (staging) as per https://github.com/brave/brave-variations/pull/957#issuecomment-1990183088.

PR #973, opened 2024-03-27
Change site visit page land cap to infinite

tmancey:
Disable site visit/page land frequency cap by changing the cap from 1 to 0 (infinite)

PR #974, opened 2024-03-27
Change site visit page land cap to infinite - Prod

tmancey:
Uplift of https://github.com/brave/brave-variations/pull/973


kjozwiak (approved):
Uplift into `production` approved 👍 QA has verified the above on `staging` via `Linux` as per https://github.com/brave/brave-variations/pull/973#issuecomment-2025340203.

PR #977, opened 2024-03-27
Disable UseWritePixelsYUV

tmancey:
Finch P5. This PR disables `UseWritePixelsYUV` between `123.*` and `125.*`.

PR #978, opened 2024-03-27
Disable feature use write pixels yuv - Prod

tmancey:
Uplift of https://github.com/brave/brave-variations/pull/977


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. I looked through https://issues.chromium.org/issues/330437904#comment32 and couldn't really find any STR/Cases other than basically getting on YT and ensuring that video/audio is still working as expected. QA went through https://github.com/brave/brave-variations/pull/977#issuecomment-2032775427 to basically ensure that YT video playback is still working once `UseWritePixelsYUV:Disabled_UseWritePixelsYUV` has been applied via Griffin.

Issue #986, opened 2024-04-03
Disable NewTabPageTriggerForPrerender2 via a kill switch


atuchin-m:
crbug: https://issues.chromium.org/issues/326509142

https://bravesoftware.slack.com/archives/C05S50MFHPE/p1711800485970059

atuchin-m:
Already disabled in the code: https://github.com/brave/brave-core/blob/1.64.x/chromium_src/chrome/browser/browser_features.cc#L25

PR #987, opened 2024-04-03
Add NewTabPageTriggerForPrerender2Killswitch

atuchin-m:
For https://github.com/brave/brave-variations/issues/986


atuchin-m:
Already disabled in the code: https://github.com/brave/brave-core/blob/1.64.x/chromium_src/chrome/browser/browser_features.cc#L25

PR #988, opened 2024-04-03
Add NewTabPageTriggerForPrerender2Killswitch

atuchin-m:
Resolves https://github.com/brave/brave-variations/issues/986
PR to main: https://github.com/brave/brave-variations/pull/987

No steps to reproduce for Brave, but it's likely result to NTP hang in some cases.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2533,6 +2533,34 @@
                     "ANDROID"
                 ]
             }
+        },
+        {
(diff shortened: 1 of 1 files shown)


atuchin-m:
Already disabled in the code: https://github.com/brave/brave-core/blob/1.64.x/chromium_src/chrome/browser/browser_features.cc#L25

PR #992, opened 2024-04-08
Move package.json to root.

goodov:
`brave-variations` should support `npm run lint/format` at the top level to allow working with study files natively.

This PR changes the location of `package.json` to allow this.


.github/workflows/test-src.yml
```diff
@@ -10,9 +10,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
```
atuchin-m, line 12: I believe have to change: `on:` trigger above. 
Otherwise package.json changes could break /src tests.

atuchin-m:
Please merge in the same time with https://github.com/brave/finch-data-private/pull/7

PR #995, opened 2024-04-09
Update ts configs, add diff-based linter.

goodov:
Sec-review to add `tsx` https://github.com/brave/reviews/issues/1575

Changes in this PR:

1. Add Prettier config for all files in the repository so it can be used natively via VSCode extension and as a part of `npm run lint`.
2. Update lint packages to most recent versions.
3. Add `lint-staged` integration to support diff linting with arbitrary executables. This will allow us to perform full study validation as a part of a single `npm run lint` call. `lint-staged` can also be integrated into `pre-commit` or `pre-push` git hooks, but after some experiments I decided to postpone this integration. 
4. Update `tsconfig` and `eslint` configs to run type checks on `src/`.


src/.eslintrc.js
```diff
@@ -2,32 +2,28 @@
...
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
+
 module.exports = {
   settings: {
     react: {
       version: 'detect',
     },
   },
   extends: ['standard-with-typescript', 'prettier', 'plugin:react/recommended'],
```
goodov, line 12: `prettier` stays here to disable `eslint` rules that `prettier` covers.

src/tsconfig.json
```diff
@@ -7,13 +7,19 @@
     "allowJs": true,
     "moduleResolution": "node",
     "baseUrl": ".",
-    "typeRoots": [
-      "../node_modules/@types",
-    ],
+    "typeRoots": ["../node_modules/@types"],
+    "jsx": "react-jsx",
```
atuchin-m, line 11: It `src/web` specific thing, isn't it? We shouldn't use React outside web.

atuchin-m (dismissed):

package.json
```diff
@@ -3,47 +3,51 @@
...
     "build:tracker": "tsc -p src/finch_tracker/tsconfig.json",
     "build:web": "webpack -c src/web/webpack.config.js --mode=production",
-    "build": "npm run build:proto && npm run build:web && npm run build:tracker",
-    "serve:dev": "webpack serve -c src/web/webpack.config.js --hot --mode=development"
+    "clean": "tsx src/scripts/clean.ts",
+    "lint": "tsx src/scripts/lint.ts",
+    "serve:dev": "webpack serve -c src/web/webpack.config.js --hot --mode=development",
+    "test": "jest --config=src/jest.config.js",
+    "tracker": "node src/finch_tracker/build/finch_tracker/main.js",
+    "typecheck": "tsc --noEmit"
```
petemill, line 15: Will this be run as a blocker for merging? Otherwise there is no type-checking for the node code.

petemill (dismissed):
I think this is a good solution, as long as `npm run typecheck` is used as a workflow to prevent merging code that fails to compile.

atuchin-m (approved):

PR #1022, opened 2024-04-25
Use json-stringify to format jsons.

goodov:
Default prettier formatter is not very good. It uses printWidth and sometimes may lay arrays in a single line and sometimes in multiple lines.

`json-stringify` in always consistent.


atuchin-m (approved):
LGTM

PR #1023, opened 2024-04-25
Sort studies by name and add basic study intersection check.

goodov:
Sort studies by name before generating seed.

This is a required change to migrate to per-file study structure, because per-file structure implies study enumeration in alphabetical order (as files lay on the file system).

This change should be a no-op as long as we don't have studies intersection. This PR adds the required validation to ensure we do not have this kind of intersection, which means if the seed is validated, then it doesn't depend on the initial order of studies in `seed.json` and it's safe to reorder studies.

Related https://github.com/brave/brave-browser/issues/33654

### DIFF
--- a/seed/serialize.py
+++ b/seed/serialize.py
@@ -138,8 +138,8 @@ def validate(seed):
             version_to_int_array(study.get('filter', {}).get('max_version')),
         ]
 
-    def is_set_intersect(a, b):
-        return a == b or a.intersection(b)
+    def is_filter_set_intersect(a, b):
+        return not a or not b or a.intersection(b)
 
     def is_version_range_intersect(range1, range2):
         return compare_versions(range1[1], range2[0]) >= 0 and compare_versions(range2[1], range1[0]) >= 0
@@ -156,9 +156,9 @@ def validate(seed):
                 study2_channel = get_study_channels(study2)
                 study2_version_range = get_study_version_range(study2)
                 # Check if the studies overlap in platform
-                if is_set_intersect(study1_platform, study2_platform):
+                if is_filter_set_intersect(study1_platform, study2_platform):
                     # Check if the studies overlap in channel
-                    if is_set_intersect(study1_channel, study2_channel):
+                    if is_filter_set_intersect(study1_channel, study2_channel):
                         # Check if the studies overlap in version
                         if is_version_range_intersect(study1_version_range, study2_version_range):
                             raise ValueError(f"Studies overlap:\n{json.dumps(study1, indent=2)}\n\n{json.dumps(study2, indent=2)}")
--- a/seed/serialize.py
+++ b/seed/serialize.py
@@ -12,6 +12,7 @@ import proto.study_pb2 as study_pb2
 import time
 import proto.variations_seed_pb2 as variations_seed_pb2
 import argparse
+import collections
 
 SEED_BIN_PATH = "./seed.bin"
 SERIALNUMBER_PATH = "./serialnumber"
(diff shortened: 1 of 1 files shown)


seed/serialize.py
```diff
@@ -213,6 +281,7 @@ def main():
 
     print("Load", args.seed_path.name)
     seed_data = json.load(args.seed_path)
+    seed_data['studies'].sort(key=lambda study: study['name'])
```
atuchin-m, line 284: I suppose python `sort` is stable?
Let's add a comment we can't reorder the study with the same name.

seed/serialize.py
```diff
@@ -44,6 +45,73 @@ def validate(seed):
...
+    def get_study_channels(study):
+        return set(study.get('filter', {}).get('channel', []))
+
+    def get_study_version_range(study):
+        def version_to_int_array(version_str):
+            parts = version_str.split('.')
+            version_list = []
+            for part in parts:
+                if part == '*':
+                    break
```
atuchin-m, line 74: Hm, I'm not 100% sure about this part.
`*` means different for min/max version.
as for Python is alway `[1,2] > [1]`

seed/serialize.py
goodov: why can't we reorder the studies with the same name if they do not overlap by filters?

atuchin-m: If they don't overlap there is no problem.
If I'm not mistaken overlapped studies with the same name is used in Yandex.

seed/serialize.py
```diff
@@ -44,6 +45,73 @@ def validate(seed):
...
+    def is_version_range_intersect(range1, range2):
+        def compare_versions(parts1, parts2):
+            if parts1 > parts2:
+                return 1
+            elif parts1 < parts2:
+                return -1
+            else:
+                return 0
+
+        return not (compare_versions(range1[1], range2[0]) < 0 or compare_versions(range2[1], range1[0]) < 0)
```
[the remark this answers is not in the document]
atuchin-m: I.e. max_version=125.* and min_version=125.1.67.1
[125] < [125,1,67,1] => the code considers there is no overlap, but in fact there is.

seed/serialize.py
goodov: > Hm, I'm not 100% sure about this part. `*` means different for min/max version. as for Python is alway `[1,2] > [1]`

yes, you're right. This comparison is tricky. Fixed and add tests from the C++ version.

seed/serialize.py
goodov: > If they don't overlap there is no problem. If I'm not mistaken overlapped studies with the same name is used in Yandex.

We should not allow overlapped studies, added checks ensures we don't.

seed/serialize.py
```diff
@@ -26,6 +27,73 @@
...
+        ["1.2.0.0.0.0", "1.2.*", 0],
+        [None, None, 0],
+        [None, "1", 0],
+        ["1", None, 0],
+    ]
+    for test_case in test_cases:
+        version1 = version_to_int_array(test_case[0])
+        version2 = version_to_int_array(test_case[1])
+        assert compare_versions(version1, version2) == test_case[2]
+
```
goodov, line 96: this is pretty dirty (having tests and logic in the same file), but we will remove this soon entirely.

seed/serialize.py
```diff
@@ -44,6 +112,57 @@ def validate(seed):
...
+    def is_set_intersect(a, b):
+        return a == b or a.intersection(b)
+
+    def is_version_range_intersect(range1, range2):
+        return compare_versions(range1[1], range2[0]) >= 0 and compare_versions(range2[1], range1[0]) >= 0
+
+    test_version_comparison()
+    for studies in feature_names_to_studies.values():
+        for i, study1 in enumerate(studies):
+            study1_platform = get_study_platforms(study1)
```
atuchin-m, line 150: Could `study_platform` or `study_channel` be empty?

goodov: > Could `study_platform` or `study_channel` be empty?

yep. improved the intersection check to cover this.

atuchin-m (approved):

PR #1024, opened 2024-04-26
Sync seed.json from production (42ce3f2583ee8f97bb76e99c332d960c643c7c2f)

goodov:
Copy `seed.json` from commit 42ce3f2583ee8f97bb76e99c332d960c643c7c2f

This is to sync-up the current state of the branches and continue migration to per-file structure in `main` branch.

Related https://github.com/brave/brave-browser/issues/33654

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -28,8 +28,7 @@
             "filter": {
                 "channel": [
                     "NIGHTLY",
-                    "BETA",
-                    "RELEASE"
+                    "BETA"
                 ],
                 "min_version": "92.1.30.57",
                 "platform": [
@@ -118,7 +117,6 @@
                 "channel": [
                     "RELEASE"
                 ],
-                "min_version": "120.0.6099.217",
                 "platform": [
                     "WINDOWS",
                     "MAC",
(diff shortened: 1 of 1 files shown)


seed/seed.json
```diff
@@ -500,7 +506,7 @@
                 {
                     "feature_association": {
                         "enable_feature": [
-                            "HangWatcher"
+                            "EnableHangWatcher"
```
atuchin-m, line 509: Thanks for fixing this.
`EnableHangWatcher` is the correct feature name.
Somehow `HangWatcher` was used in `main` by mistake.

seed/seed.json
```diff
@@ -666,26 +671,25 @@
...
+                "min_version": "103.1.43.9",
                 "platform": [
                     "WINDOWS",
                     "MAC",
-                    "LINUX",
-                    "ANDROID"
+                    "LINUX"
                 ]
             },
-            "name": "BraveTranslateStudy"
```
atuchin-m, line 678: I've checked that both translate-related features `BraveTranslateStudy` and `BraveTranslateStudy` are same before/after the PR. The only things is changed here is the order.

BTW, both features are outdated and limited by `max_version`

PR #1026, opened 2024-04-26
Add seed tools.

goodov:
This PR adds `seed_tools` CLI tool to support seed manipulation stored in per-file format.

Related: https://github.com/brave/brave-browser/issues/33654


src/seed_tools/utils/seed_validation.ts
```diff
@@ -0,0 +1,260 @@
...
+      doesFilterArrayIntersect(study1.filter?.channel, study2.filter?.channel),
+    platform: (study1: Study, study2: Study) =>
+      doesFilterArrayIntersect(
+        study1.filter?.platform,
+        study2.filter?.platform,
+      ),
+    locale: (study1: Study, study2: Study) =>
+      doesFilterArrayIntersect(study1.filter?.locale, study2.filter?.locale),
+    exclude_locale: (study1: Study, study2: Study) =>
+      doesFilterArrayIntersect(
```
atuchin-m, line 96: I'm not sure about `exclude_*` logic..
If we have study A with `exclude_locale: ["en"]` and study B with `exclude_locale: ["es"]`, they actually intersect (e.g. if user locale is `fr`) .

src/seed_tools/utils/version.ts
```diff
@@ -0,0 +1,138 @@
...
+      (this.components.length > 0 && this.isWildcard ? '.*' : '')
+    );
+  }
+}
+
+// Strictly parse a string as a number that can be represented as uint32. This
+// is required to match the behavior of base::Version in C++.
+function parseVersionComponent(str: string): number {
+  if (!/^\d+$/.test(str)) {
+    throw new Error(
```
atuchin-m, line 126: minor: maybe return `undefined` here, and throw an error right after to show the full version with an issue (not only the bad component)?

src/seed_tools/utils/version.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      {
+        input: '1.2.3.4.5.6.7.8.9.0',
+        parts: 10,
+        firstpart: 1,
+        success: true,
+      },
+      { input: '02.1', parts: 0, firstpart: 0, success: false },
+      { input: '0.01', parts: 2, firstpart: 0, success: true },
+      { input: 'f.1', parts: 0, firstpart: 0, success: false },
+      { input: '15.007.20011', parts: 3, firstpart: 15, success: true },
```
atuchin-m, line 45: what about replacing `...parts: 0, firstpart: 0, success: false ` to `parts: [15, 007, 20011]`?
(`parts: Array<str | number> | undefined`)

src/seed_tools/utils/version.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    testCases.forEach((testCase) => {
+      it(`should compare ${testCase.lhs} and ${testCase.rhs}`, () => {
+        const lhs = new Version(testCase.lhs);
+        const rhs = new Version(testCase.rhs);
+        expectCompare(lhs, rhs, testCase.expected);
+      });
+    });
+
+    describe('wildcard', () => {
+      const testCases = [
```
atuchin-m, line 181: nit: if we list test cases in ascending order we could just compare each pair, don't need to set specify `expected`

src/seed_tools/utils/study_json_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function jsonStudyReplacer(key: string, value: any): any {
+  switch (key) {
+    case 'start_date':
+    case 'end_date': {
+      const intValue = Number.isInteger(value) ? value : parseInt(value);
+      return new Date(intValue * 1000).toISOString();
+    }
+    case 'channel':
```
atuchin-m, line 83: Could we pass a param isBrave: boolean = true to make the tools work for Finch seed?

src/seed_tools/utils/version.test.ts
goodov: kept this as is. Most of these tests are direct conversion from `base/version_unittest.cc`.

goodov:
@atuchin-m @iefremov ptal

src/seed_tools/utils/diff_strings.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        const result = stdout
+          .replaceAll(tmpFile1, displayFileName1)
+          .replaceAll(tmpFile2, displayFileName2);
+
+        resolve(result);
+      },
+    );
+  }).finally(() => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
atuchin-m, line 50: Maybe just add `.catch(console.error)` to remove this supression?

src/seed_tools/utils/diff_strings.ts
```diff
@@ -0,0 +1,56 @@
...
+    .replaceAll('\\', '/');
+  const tmpFile2 = path
+    .join(tmpDir, `file2-${Date.now()}.txt`)
+    .replaceAll('\\', '/');
+
+  // Write strings to temporary files.
+  await fs.writeFile(tmpFile1, string1);
+  await fs.writeFile(tmpFile2, string2);
+
+  return await new Promise<string>((resolve, reject) => {
```
atuchin-m, line 30: nit: remove await.

src/seed_tools/utils/diff_strings.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Use git diff to generate readable diff.
+    exec(
```
atuchin-m, line 32: Why not `execa.execaSync`?
Or `util.promisify` over `exec()`?

src/seed_tools/utils/seed_validation.ts
```diff
@@ -0,0 +1,304 @@
...
+    (study1Filter.length === 0 && study1ExcludeFilter.length === 0) ||
+    (study2Filter.length === 0 && study2ExcludeFilter.length === 0)
+  ) {
+    return true;
+  }
+
+  // If both studies have include filters, they intersect if at least one value
+  // is in common.
+  if (study1Filter.length > 0 && study2Filter.length > 0) {
+    return study1Filter.some((locale) => study2Filter.includes(locale));
```
atuchin-m, line 268: `s/locale/value` ?

src/base/containers/default_map.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2024 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+export type DefaultFactory<K, V> = (key?: K) => V;
```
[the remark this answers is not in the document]
goodov: technically we don't, but it's used in the public interface. It's better to export such things.

src/base/result.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2024 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import assert from 'assert';
+
+export default class Result<T, E> {
+  private readonly _value: T | undefined;
+  private readonly _error: E | undefined;
```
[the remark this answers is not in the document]
goodov: This way you need to `import { type Ok } from ...` everywhere you want to use the function call result. I don't like it. Can we avoid this?

src/scripts/generate_proto_ts.ts
```diff
@@ -0,0 +1,123 @@
...
+  );
+}
+
+function generateProtobufTs() {
+  try {
+    // Apply study.proto patch to make protobuf-ts serialize probability_weight
+    // field if its value set to 0.
+    gitApplyStudyProtoPatch();
+    execa.execaSync(
+      'protoc',
```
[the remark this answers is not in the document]
goodov: no, it's as a part of `protobuf-ts/plugin`.
https://github.com/timostamm/protobuf-ts/blob/main/packages/protoc/README.md

src/seed_tools/utils/diff_strings.ts
goodov: eslint disagrees.
```
30:10  error  Returning an awaited promise is required in this context  @typescript-eslint/return-await
```

src/seed_tools/utils/diff_strings.ts
goodov: sadly, `execa` doesn't work in tests (esm/non-esm stuff).

replaced with `promisify(exec)`.

src/seed_tools/utils/diff_strings.ts
atuchin-m: ok)

src/seed_tools/utils/seed_validation.ts
```diff
@@ -0,0 +1,300 @@
...
+              false
+            );
+          },
+        );
+
+        // If there is no filter difference, the studies overlap in all
+        // properties, meaning the same feature is used in both.
+        if (!hasFilterDifference) {
+          errors.push(
+            `Feature ${featureName} overlaps in studies. Check your filters:\n${study_json_utils.stringifyStudyArray([study1.study, study2.study])}`,
```
atuchin-m, line 114: nit: a long line, better to split

atuchin-m (approved):

src/base/result.ts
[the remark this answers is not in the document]
goodov: replaced with even simpler approach where we can check for `instaneof Error` which is always defined ;)
thanks!

src/seed_tools/utils/diff_strings.ts
goodov: removed `execa` entirely, as there's no point in keeping it. I was hoping to use it everywhere instead of `child_process`, but the fact that it cannot be used in tests (import just doesn't work), means we should not spend time with it and just use the good ol' `child_process`..

atuchin-m: thanks.
It also good in terms of security/

Issue #983, opened 2024-04-02
Hide aggressive fingerprinting mode in Release


arthuredelstein:
Follow on to #966

arthuredelstein:
Done!

PR #1038, opened 2024-04-30
hide strict fingerprinting mode for 50% of Release clients

arthuredelstein:
Progress in #983


kjozwiak (approved):

PR #1052, opened 2024-05-07
Sync HangWatcher main with production

atuchin-m:
Fixes a conflict `main` vs `production`. 
production is a correct one: https://github.com/brave/brave-variations/blob/production/seed/seed.json#L509


kjozwiak (approved):
LGTM 👍 Merging into `main`.

Issue #1060, opened 2024-05-16
Disable Canvas2DHibernation via a kill switch


atuchin-m:
crbug: https://issues.chromium.org/issues/328755781
https://source.chromium.org/chromium/chromium/src/+/refs/tags/125.0.6422.60:third_party/blink/common/features.cc;l=515-518

PR #1061, opened 2024-05-16
Disable Canvas2DHibernation via a kill switch

atuchin-m:
For https://github.com/brave/brave-variations/issues/1060


kjozwiak (approved):
LGTM 👍 Approving so we can verify the above on `staging` before uplifting https://github.com/brave/brave-variations/pull/1063 into `production`.

PR #1076, opened 2024-05-31
Added BraveSearchPromotionDDGBannerStudy

simonhong:
Added to another search promotion targets for DDG users
We targets 100% users and each type has 25%.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -820,6 +820,36 @@
                 {
                     "feature_association": {
                         "enable_feature": [
+                            "BraveSearchOmniboxDDGBanner"
+                        ]
+                    },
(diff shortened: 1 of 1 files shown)


simonhong:
@rebron this doesn't have min version as I'm not sure it's started from 1.68.x or not.
If we want to start this from 1.67.x, I'll set 1.67 as min version. Otherwise, it could be 1.68.

kjozwiak (commented):
Going to close the above as discussed with @rebron via the desktop call. It's been opened for ~2 months without any movement. I believe the `Search Promotion` feature is also being changed/re-worked. The DDG images/campaign is also being re-worked/needs new assets.

PR #1083, opened 2024-06-06
Add Github actions linter

goodov:
Add actions linter, because github does not notify if an action is broken, so it's possible to merge invalid action without noticing.

### DIFF
--- a/.github/workflows/validate-actions.yml
+++ b/.github/workflows/validate-actions.yml
@@ -1,0 +1,20 @@
+name: Lint GitHub Actions
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/validate-actions.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
+
+      - name: Check workflow files
+        run: |
+          echo "::add-matcher::.github/matchers/actionlint-matcher.json"
+          docker run --rm -v "$PWD:/repo:ro" --workdir /repo rhysd/actionlint:1.7.1 -color
```
goodov, line 19: using docker here to fix the version and have everything actionlint might require be built-in.

downloading the linter via the script they offer is not perfect, as it can only reference github tags.

mihaiplesa (approved):

PR #1088, opened 2024-06-12
Add Griffin study for BraveWalletAnkrBalances feature

onyb:
Adding a study for the `BraveWalletAnkrBalances` feature enabling it for 50% of the users. The flag is available as of **1.62.x** and is disabled by default.

PR #1095, opened 2024-06-19
Remove Android from Ankr study

onyb:
A study for `BraveWalletAnkrBalances` was added in https://github.com/brave/brave-variations/pull/1088, which included Android as one of the target platforms. However, due to some [issues](https://github.com/brave/brave-variations/pull/1088#issuecomment-2176599096) encountered in Android during QA, we are going ahead with the rollout of the study without Android.

This should continue to the be case until the Android issue has been resolved.

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2977,8 +2977,7 @@
                 "platform": [
                     "WINDOWS",
                     "MAC",
-                    "LINUX",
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
LGTM 👍Merging.

PR #1109, opened 2024-07-03
Rollout P3A Nebula to 10% of release users

DJAndries:
To be merged after the release of 1.68

PR #1110, opened 2024-07-03
Rollout P3A Nebula to 10% of release users [production]

DJAndries:
To be merged after the release of 1.68

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2915,6 +2915,36 @@
                 ]
             },
             "name": "DefaultPlaylistStudy"
+        },
+        {
+            "experiments": [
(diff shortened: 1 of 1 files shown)


kjozwiak:
We'll push the above into `production` once `1.68.x` hits production as it will include https://github.com/brave/brave-browser/issues/35841.

kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the above on `Release` using `main (staging)` as per https://github.com/brave/brave-variations/pull/1109#issuecomment-2256768570.

PR #1111, opened 2024-07-03
Deprecate all JSON P3A measurements for non-macOS/iOS platforms

DJAndries:
To validate:

- Ensure no requests are sent to `p3a-json.brave.com` for non-macOS/iOS platforms
- Ensure requests are still sent to `p3a-json.brave.com` for macOS and iOS
- Ensure requests are sent to `star-randsrv.bsg.brave.com`, `collector.bsg.brave.com/typical`, `collector.bsg.brave.com/express`, `collector.bsg.brave.com/slow` and `collector.bsg.brave.com/creative` for all platforms

PR #1112, opened 2024-07-03
Deprecate all JSON P3A measurements for non-macOS/iOS platforms [production]

DJAndries:
See https://github.com/brave/brave-variations/pull/1111 for validation steps


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR on `staging` using `Release` as per https://github.com/brave/brave-variations/pull/1111#issuecomment-2218334682

Issue #1114, opened 2024-07-08
Set NoRecency Leg of BraveAds.CreativeAdModelBasedPredictorRecencyStudy to 100%


ptjames:
This issue is for updating most performant test leg of ad model predictor recency study to 100%.

PR #1115, opened 2024-07-08
Updating most performant test leg of ad model predictor recency study

ptjames:
Updating most performant test leg of ad model predictor recency study to 100% until brave-core changes reach production.

Issue: https://github.com/brave/brave-variations/issues/1114
Changes become permanent in brave-core via this PR: https://github.com/brave/brave-core/pull/24253

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2351,28 +2351,7 @@
                             "value": "0.0"
                         }
                     ],
-                    "probability_weight": 20
-                },
-                {
-                    "feature_association": {
-                        "enable_feature": [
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved 👍 As discussed via https://bravesoftware.slack.com/archives/C01LKMP6X36/p1720454503659199, we're just increasing the % so there's nothing that QA needs to really do in terms of verification. Has been previously checked via https://bravesoftware.slack.com/archives/CHGKGMHDJ/p1716481614558189. This will also be enabled `100%` by default once https://github.com/brave/brave-core/pull/24253 makes it into `production`.

PR #1117, opened 2024-07-09
Add test seed upload workflow.

goodov:
Add workflows for upload and removal of test seed.
Add a single study to test these workflows.


.github/workflows/upload-test-seed.yml
```diff
@@ -0,0 +1,94 @@
...
+
+            ### What's Next?
+
+            - The upload process typically takes a few minutes.
+            - Once the upload is complete, this comment will provide the seed details and testing instructions.
+            `
+            const comment = require('.github/workflows/scripts/comment.js')
+            await comment(github, context, commentBody)
+
+      - name: Install dependencies and run tests
```
atuchin-m, line 34: nit: could we split it into?
1. npm install
2. build proto & typecheck
3. tests

.github/workflows/upload-test-seed.yml
```diff
@@ -0,0 +1,94 @@
...
+            \`\`\`
+
+            #### Seed Details
+            - Serial Number: \`${serialNumberContent}\`
+            - Uploaded: \`${new Date().toISOString()}\`
+            `
+            const comment = require('.github/workflows/scripts/comment.js')
+            await comment(github, context, commentBody)
+
+      - name: Comment "Upload Failed"
```
atuchin-m, line 82: I would prefer to move this up and rename to "Upload skipped due to a build/lint/test failure".

Not sure that we should make a fallback for s3 uploading / invalidation, because it's a very small chance of failure here. IMHO, if it happens it's enough to have "Upload In Progress" comment and the failed job in the list.

.github/workflows/delete-test-seed.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true
+    env:
+      REMOTE_SEED_PATH: 'pull/${{ github.event.pull_request.number }}/seed'
+
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
```
atuchin-m, line 17: nit: do we need this?

.github/workflows/delete-test-seed.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  pull_request:
+    types: [closed]
+    paths:
+      - 'studies/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true
```
atuchin-m, line 12: what if the PR is declined?

goodov: ah, we don't need this check at all.

atuchin-m (approved):

PR #1143, opened 2024-07-25
Enabled ModuleFileNamePatch on Release channel.

boocmp:

PR #1144, opened 2024-07-25
Enabled ModuleFileNamePatch on Release channel.

boocmp:


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the above on `Release` using `main (staging)` as per https://github.com/brave/brave-variations/pull/1143#issuecomment-2256798244.

PR #1146, opened 2024-07-30
Disable BraveWalletAnkrBalances rollout

onyb:
Too many operational issues on production, so we have decided to disable rollout of Ankr balances feature until there's consistent evidence of better performance.

PR #1152, opened 2024-08-02
Disable BraveWalletAnkrBalances rollout

goodov:
Cherry-pick https://github.com/brave/brave-variations/pull/1146

cc @onyb

### DIFF
--- a/seed/seed.json
+++ b/seed/seed.json
@@ -2832,11 +2832,11 @@
                         ]
                     },
                     "name": "Enabled",
-                    "probability_weight": 50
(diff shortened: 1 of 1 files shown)


kjozwiak (approved):
Uplift into `production` approved after deliberating with @brave/uplift-approvers. QA has verified the PR using `Linux` on `staging` as per https://github.com/brave/brave-variations/pull/1146#issuecomment-2262782389. We're disabling the feature due to several operational issues. It will be reenabled once performance is stabilized/improved.

PR #1156, opened 2024-08-06
Update rollout percentage for `DayZeroStudyWindows`

DJAndries:


seed/seed.json
```diff
@@ -2618,7 +2618,7 @@
                         }
                     ],
                     "name": "EnabledA",
-                    "probability_weight": 4
+                    "probability_weight": 5
```
goodov, line 2621: be aware that by doing this you're reassigning 1% of users from `variant:b` to `variant:a`. For a/b testing this might not be a desired outcome.

The correct way of doing such a/b testing is to create preperiod groups (this is how Chromium calls them) with a default state that you will then use to introduce users to a desired state. Like so:
```
initial state:

Default_Preperiod_A: 50%
Default_Preperiod_B: 50%
```

first update:
```
EnabledA: 4%
Default_Preperiod_A: 46%
EnabledB: 4%
Default_Preperiod_B: 46%
```

second update:
```
EnabledA: 5%
Default_Preperiod_A: 45%
EnabledB: 5%
Default_Preperiod_B: 45%
```

etc.

DJAndries: thanks for pointing this out. is there any documentation on preperiod groups? this was the only reference that i could find: https://source.chromium.org/chromium/chromium/src/+/main:components/safe_browsing/core/browser/user_population.cc;drc=c571efbef3ecfe66613d524ff39c58e25d18548c;l=92

i'm guessing that increasing the percentage of one of the experiment groups will ensure that new users included in the study are pulled from the preperiod groups only. is that right?

am i required to create a new study in order to utilize this functionality? if so, i assume that would mean that we would have to reset everyone's state for this feature, but only once. is that assumption correct?

goodov: > is there any documentation on preperiod groups?

no, there's no any doc on this, but it's just a name for the pre-allocated group in the study, the name doesn't matter.

> i'm guessing that increasing the percentage of one of the experiment groups will ensure that additional users will be pulled from the preperiod groups only. is that right? why do we need two preperiod groups?

We had to use them because of how Variations work in Chromium: **Variations Service does not store the "random dice value" for each new study**, instead the random value is determined from the global `randomization_seed` (bound to the browser installation), and the study name [1].

When the browser parses a study, it enumerates experiments in the order they are listed. When the `probability_weight` sum crosses the "random dice value" for this study, the browsers makes the group active.

By pre-allocating such groups you allow yourself to chip away a userbase percentage from a non-modified group into a tested group in a controllable way.

1. https://chromium.googlesource.com/chromium/src/+/129.0.6626.1/base/metrics/field_trial.h#140

goodov (approved):

kjozwiak (approved):

PR #1159, opened 2024-08-06
Removed BgPlaybackStudy on Android

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/1158

PR #1160, opened 2024-08-06
Removed BgPlaybackStudy on Android

SergeyZhukovsky:
Production version of https://github.com/brave/brave-variations/pull/1159


kjozwiak (commented):
Closing the above as we don't need `production` PRs anymore now that https://github.com/brave/devops/pull/12117 was merged and the `production` branch is read only.

PR #1169, opened 2024-08-09
Update Griffin web UI to use "MAIN" instead of "PRODUCTION".

goodov:


src/core/base_types.ts
```diff
@@ -4,8 +4,8 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 export enum SeedType {
-  PRODUCTION, // production Brave seed (brave-variations@production)
-  STAGING, // staging Brave seed (brave-variations@main)
+  MAIN, // main Brave seed (brave-variations@main)
+  STAGING, // staging Brave seed (brave-variations@staging)
```
atuchin-m, line 8: IMHO we don't need to support `staging` here in the long term perspective.

atuchin-m (approved):

Issue #1168, opened 2024-08-09
Support seed_tools (with multiple files) in fieldtrials_testing_config_generator.py


atuchin-m:
The old generator will be deprecated soon.

PR #1171, opened 2024-08-09
Use seed_tools in fieldtrials_testing_config_generator.py

atuchin-m:
Resolves https://github.com/brave/brave-variations/issues/1168

### DIFF
--- a/seed/fieldtrials_testing_config_generator.py
+++ b/seed/fieldtrials_testing_config_generator.py
@@ -17,17 +17,19 @@ https://chromium.googlesource.com/chromium/src/+/master/testing/variations/READM
 """
 
 import argparse
-import serialize
 import json
-import proto.study_pb2 as study_pb2
+import os
+import re
+import serialize
 import subprocess
 import sys
-import proto.variations_seed_pb2 as variations_seed_pb2
-import re
+import tempfile
 
 from datetime import datetime, timezone
 from packaging import version
 
+import proto.study_pb2 as study_pb2
+import proto.variations_seed_pb2 as variations_seed_pb2
 
 PLATFORM_NAMES = {
     study_pb2.Study.Platform.PLATFORM_WINDOWS: 'windows',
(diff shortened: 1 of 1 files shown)


goodov:
let's add CI test for this. I think the test should just use the current commit and validate that it can generate the required json file.

seed/fieldtrials_testing_config_generator.py
```diff
@@ -50,10 +54,48 @@ def _get_variations_revision(date: str, branch: str) -> str:
     return output.rstrip().decode('utf-8')
 
 
-def _get_seed_data(seed_git_path: str, variations_revision: str):
+def _get_variations_seed_legacy(revision: str):
     seed_string = subprocess.check_output(
```
goodov, line 58: this will raise if the file is not found.

goodov: you won't get to the non-legacy part of the code seed generator if the exception is thrown here.

atuchin-m: fixed.

atuchin-m:
> let's add CI test for this. I think the test should just use the current commit and validate that it can generate the required json file.

I've added `python3 seed/fieldtrials_testing_config_generator.py` to `test.yml`

.github/workflows/test.yml
```diff
@@ -25,6 +25,13 @@ jobs:
       - name: Serialize seed
         run: python seed/serialize.py seed/seed.json
 
+      - name: Generate testing_fieldtrials.json for perf tests
+        run: python3 seed/fieldtrials_testing_config_generator.py
+          --output fieldtrial_testing_config.json
+          --output-revision brave-variations-revision.txt
+          --target-date "2024-08-12 00:02:25 +0000"
```
goodov, line 32: this is not enough. the test should also run on the current revision. pls update the script to do this, i.e. instead of `--target-*` flags there can be `--commit` flag.

seed/fieldtrials_testing_config_generator.py
```diff
@@ -142,22 +192,19 @@ def main():
             '"Format: "2022-09-09 10:02:27 +0000"'))
     args = parser.parse_args()
 
-    date = datetime.strptime(args.target_date, '%Y-%m-%d %H:%M:%S %z')
+    if args.target_date:
+        date = datetime.strptime(args.target_date, '%Y-%m-%d %H:%M:%S %z')
+    else:
+        date = datetime.now(tz=timezone.utc)
```
goodov, line 198: this will use the `main` branch, not the current branch/commit.

the test should validate changes made in the PR.

goodov (approved):

seed/fieldtrials_testing_config_generator.py
atuchin-m: updated

PR #1176, opened 2024-08-09
Fix codeowners order.

goodov:
RTFM 🤦 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,8 @@
 # The order is important. The last matching pattern takes the most precedence.
 
+# Griffin maintainers are responsible for all machinery in this repository.
 * @brave/griffin-maintainers
 
+# The seed file is owned by the uplift team as any change to it directly affects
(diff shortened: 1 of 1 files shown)


mihaiplesa (approved):

.github/CODEOWNERS
```diff
@@ -1,2 +1,4 @@
-seed/seed.json   @brave/uplift-approvers
+# The order is important. The last matching pattern takes the most precedence.
+
 *                @brave/griffin-maintainers
+/seed/seed.json   @brave/uplift-approvers
```
mihaiplesa, line 4: Shouldn't the @ come left by 1 space?

goodov: yes, it looked wrong. Instead I've added comments and separated entries, so we don't need to sync offsets on all lines at all.

mihaiplesa (approved):

PR #1183, opened 2024-08-19
Add more study validation from processed_study.cc.

goodov:
Implement additional validation that exists in [processed_study.cc](https://source.chromium.org/chromium/chromium/src/+/main:components/variations/processed_study.cc;l=38;drc=f9d386e085b36b0118ab74f9a3d6dc18934de430).

Important ones:
* study, experiment, feature name validation
* default_experiment_name validation + experiment existence validation
* experiment param name conflict and emptiness

Other was added just to have the full coverage (forcing_flag, google_web_experiment_id).


src/seed_tools/utils/study_validation.ts
```diff
@@ -80,11 +88,97 @@ function checkExperiments(study: Study): string[] {
...
+    study.default_experiment_name !== '' &&
+    !experimentNames.has(study.default_experiment_name)
+  ) {
+    errors.push(
+      `Missing default experiment: ${study.default_experiment_name} in ${study.name} study`,
+    );
+  }
+
+  // Validate total probability.
   if (totalProbability !== 100) {
```
iefremov, line 178: generally it's not neccessarily true, but probably we can limit ourself to 100 for simplicity
image

iefremov (approved):

src/seed_tools/utils/study_validation.ts
```diff
@@ -80,11 +88,97 @@ function checkExperiments(study: Study): string[] {
     }
     experimentNames.add(experiment.name);
     totalProbability += experiment.probability_weight ?? 0;
+
+    // Validate features.
+    if (experiment.feature_association !== undefined) {
+      if (
+        experiment.feature_association.enable_feature !== undefined &&
```
atuchin-m, line 95: Can we use `experiment?.feature_association` to remove ` if (experiment.feature_association !== undefined)`?
Also, `!==undefined` can't be get rid off if it's an object.

src/seed_tools/utils/study_validation.ts
```diff
@@ -161,3 +255,48 @@ function checkOsVersionRange(study: Study): string[] {
...
+  }
+}
+
+function isStringASCIIWithoutChars(
+  str: string,
+  charsToExclude: string,
+): boolean {
+  // Check if every character in the string is within the ASCII range (0-127).
+  for (let i = 0; i < str.length; i++) {
+    if (str.charCodeAt(i) > 127) {
```
atuchin-m, line 294: Do we need to check <256?

src/seed_tools/utils/study_validation.ts
```diff
@@ -161,3 +255,48 @@ function checkOsVersionRange(study: Study): string[] {
...
+    );
+  }
+}
+
+function isStringASCIIWithoutChars(
+  str: string,
+  charsToExclude: string,
+): boolean {
+  // Check if every character in the string is within the ASCII range (0-127).
+  for (let i = 0; i < str.length; i++) {
```
atuchin-m, line 293:
```suggestion
  for (const i of str) {
```

src/seed_tools/utils/study_validation.ts
```diff
@@ -80,11 +88,97 @@ function checkExperiments(study: Study): string[] {
...
     experimentNames.add(experiment.name);
     totalProbability += experiment.probability_weight ?? 0;
+
+    // Validate features.
+    if (experiment.feature_association !== undefined) {
+      if (
+        experiment.feature_association.enable_feature !== undefined &&
+        experiment.feature_association.enable_feature.length > 0
+      ) {
+        for (const feature of experiment.feature_association.enable_feature) {
```
atuchin-m, line 95: WDYT about making `let features_to_check: string[] = []` and extending it.
Then just a one for-loop to check them all.

src/seed_tools/utils/study_validation.ts
goodov: > We can assume `totalProbability == 100` for Griffin

yes, that's what we did before and I assume we will continue doing. It's just easier to view this as 0-100 percent value.

src/seed_tools/utils/study_validation.ts
```diff
@@ -80,11 +88,79 @@ function checkExperiments(study: Study): string[] {
...
+    ) {
+      errors.push(
+        `Experiment ${experiment.name} has both google_web_experiment_id and web_trigger_experiment_id`,
+      );
+    }
+
+    // Valiate params.
+    const paramNames = new Set<string>();
+    if (experiment.param.length > 0) {
+      for (const param of experiment.param) {
```
atuchin-m, line 133: nit: `for (const param of experiment.param || [])`

atuchin-m (approved):

atuchin-m (approved):

PR #1184, opened 2024-08-20
Use @typescript-eslint/recommended-type-checked for lint

atuchin-m:
`standard-with-typescript` [is deprecated](https://github.com/brave/brave-variations/issues/961).
A config is used: https://typescript-eslint.io/users/configs#projects-with-type-checking

`package-lock.json` was recreated (removed, `npm install`, `git add` )


package.json
```diff
@@ -46,7 +44,8 @@
     "ts-jest": "29.2.4",
     "ts-loader": "9.5.1",
     "tsx": "4.17.0",
-    "typescript": "5.5.4",
+    "typescript": "^5.5.4",
```
goodov, line 47: pls make all version be exact (without `^` symbol). 

`npm config set save-exact=true` and reinstall these packages.

goodov (approved):

PR #1200, opened 2024-08-26
Increase WebcompatExceptionsService in Release to 100%

arthuredelstein:


kjozwiak (approved):

goodov:
@arthuredelstein please enable the feature by default in brave-core. Ideally we should always do this before rolling the feature to Release channel.

PR #1243, opened 2024-10-28
Switch perf to studies

atuchin-m:


goodov (approved):

.github/workflows/test.yml
```diff
@@ -23,9 +24,6 @@ jobs:
       - name: Install requirements
         run: pip install -r seed/requirements.txt
 
-      - name: Serialize seed
-        run: python seed/serialize.py seed/seed.json
-
       - name: Generate testing_fieldtrials.json (HEAD)
```
goodov, line 27: let's rename this workflow file to be something like `test-testing-fieldtrials-gen.yml`, and also rename `name: Test Python Generator` to a better suitable name.

.github/workflows/test.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - 'seed/**'
+      - 'studies/**'
```
goodov, line 7: let's also add `- 'src/**'`

goodov (approved):

.github/workflows/test.yml
atuchin-m: done.

.github/workflows/test.yml
atuchin-m: added.

goodov (approved):

PR #1253, opened 2024-11-06
Unify proto JS API for tracker and griffin UI

atuchin-m:
The PR reimplements finch_tracker and griffin.brave.com using `protobuf-ts` instead `protobufjs` 
Also it reuses the serializer from seed_tools that will result in changing the format in finch-data-private


src/finch_tracker/tracker_lib.ts
```diff
@@ -23,18 +24,17 @@ export async function fetchChromeSeedData(): Promise<Buffer> {
...
-    const json = studyToJSON(study);
+): Record<string, Study[]> {
+  const map: Record<string, Study[]> = {};
+  const seed = VariationsSeed.fromBinary(seedData);
+  const addStudy = (path: string, study: Study) => {
     const list = map[path];
-    if (list !== undefined) list.push(json);
-    else map[path] = [json];
+    if (list !== undefined) list.push(study);
+    else map[path] = [study];
```
goodov, line 37: not a fan of these if/else one-liners. Would like these to always have a scoped body `{ ... }` similar to what clang-format now does.

src/seed_tools/utils/study_json_utils.ts
```diff
@@ -50,7 +51,7 @@ export function parseStudies(
 ): Study[] {
   const jsonStudies = JSON5.parse(
     studyArrayString,
-    jsonStudyReviever.bind(null, options),
+    jsonStudyReviewer.bind(null, options),
```
goodov, line 54: it should actually be `jsonStudyReviver` 😄 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#reviver

src/seed_tools/utils/study_json_utils.ts
```diff
@@ -96,32 +97,20 @@ function jsonStudyReplacer(
...
-      return value.map((value: string): string => {
-        return value.replace(/^PLATFORM_/, '');
-      });
+    case 'channel': {
+      return value.map((c: string) =>
+        channelToString(c, options?.isChromium === true),
+      );
+    }
+    case 'platform': {
+      return value.map((p: string) => platformToString(p));
```
goodov, line 106: before this change the logic for revive and replace was in the same file close to each other. Now it's spread a bit. Can we do something here?

src/seed_tools/utils/study_json_utils.test.ts
```diff
@@ -112,7 +112,7 @@ describe('stringifyStudies', () => {
         filter: {
           start_date: Math.floor(startDate.getTime() / 1000),
           channel: ['CANARY', 'BETA', 'STABLE'],
-          platform: ['PLATFORM_LINUX', 'PLATFORM_MAC'],
+          platform: ['LINUX', 'MAC'],
```
goodov, line 115: should the test be renamed now?

src/core/url_utils.ts
```diff
@@ -51,8 +51,7 @@ export function getStudyRawConfigUrl(
   return (
     'https://github.com/search?type=code' +
     '&q=repo%3Abrave%2Fbrave-variations' +
-    '+path%3A%2F%5Eseed%5C%2Fseed.json%7C%5Estudies%5C%2F*.json5%2F+' +
-    `"%5C"name%5C"%3A+%5C"${encodeURIComponent(study)}%5C""`
+    `+path%3Astudies%2F+%22name%3A+%27${encodeURIComponent(study)}%27%22`
```
goodov, line 54: no need for `.json5` here?

src/finch_tracker/tracker_lib.ts
atuchin-m: agreed, this `else` isn't not the case when a one-liner looks good.
Added {}

src/seed_tools/utils/study_json_utils.ts
atuchin-m: O_o. Reverted back

src/seed_tools/utils/study_json_utils.test.ts
atuchin-m: I've renamed the test.

src/core/url_utils.ts
atuchin-m: We don't need it here, because it's a code pattern to look in brave-variations. 
But we need to update the URL above.

src/seed_tools/utils/study_json_utils.ts
goodov: no, I had a typo here actually, let's fix it: `jsonStudyReviever` -> `jsonStudyReviver`.

src/seed_tools/utils/study_json_utils.ts
atuchin-m: https://github.com/brave/brave-variations/pull/1253/commits/5ef0b4e95c04499b7c0d4b55fdbaa9c382df5ae5

src/seed_tools/utils/study_json_utils.ts
atuchin-m: Fixed ;)

goodov (approved):

PR #1259, opened 2024-11-11
Rollout Llama as default model to 100% of users

nvonpentz:
Stats look solid at 75%, let's roll it out for everyone.


kjozwiak (approved):

PR #1279, opened 2024-12-09
Remove Day Zero study

DJAndries:


kjozwiak (approved):
Uplift into `main/production` has been approved after deliberating with @brave/uplift-approvers.

PR #1284, opened 2024-12-18
Duplicate countries to be in lowercase for x-country header update.

goodov:
List countries both in upper and lowercase so we can safely update server to use lowercase value in x-country header.

Related devops issue https://github.com/brave/devops/issues/12890


src/seed_tools/utils/study_validation.ts
```diff
@@ -343,3 +344,27 @@ export function validateName(
...
+function checkCountry(study: Study): string[] {
+  const errors: string[] = [];
+  if (study.filter === undefined) {
+    return errors;
+  }
+
+  // check if country in uppercase also exists in lowercase
+  const checkCountriesCase = (countries: string[]) => {
+    const countrySet = new Set(countries);
+    for (const country of countries) {
```
atuchin-m, line 357: minor: `countries.forEach((country) => {...})`

src/seed_tools/utils/studies_to_seed.ts
```diff
@@ -190,3 +194,26 @@ function setStudyDefaultParameters(study: Study) {
     study.consistency = Study_Consistency.PERMANENT;
   }
 }
+
+function duplicateCountriesAsLowercase(variationsSeed: VariationsSeed) {
+  // For now duplicate the country filters to lowercase.
+  const duplicate = (countries: string[]) => {
```
atuchin-m, line 200: What about `Array.from(new Set([...countries, ...countries.map((c) => c.toLowerCase())]))`?

atuchin-m (approved):
LGTM with a few nits.

src/seed_tools/utils/studies_to_seed.ts
goodov: this will change the order of countries

src/seed_tools/utils/study_validation.ts
goodov: seems like a style preference to me. I prefer the usual `for (...) {}` syntax similar to what we have in C++ codebase in Chromium.

Issue #1293, opened 2025-01-13
Adblock Custom Scriptlets in Nightly & Beta .


boocmp:
Enable `CosmeticFilteringCustomScriptlets` feature in Nightly and Beta

PR #1294, opened 2025-01-13
Added AdblockCustomScriptlets study.

boocmp:
Resolves https://github.com/brave/brave-variations/issues/1293

Original issue: https://github.com/brave/brave-core/pull/25999


boocmp:
We decided to enable the corresponding feature by default.

PR #1314, opened 2025-02-25
Set new menu rollout on iOS to 50%

kylehickinson:
1.75 is fully rolled out now and we haven't heard any negative feedback on the new menu yet


kjozwiak (approved):

PR #1317, opened 2025-02-27
Set new menu rollout on iOS to 100%

kylehickinson:


kjozwiak (approved):

PR #1320, opened 2025-03-04
Disable KeywordTableHashVerification.

boocmp:
I've looked in https://chromium.googlesource.com/chromium/src/+/refs/tags/133.0.6943.141/components/search_engines/keyword_table.cc#39, this feature is enabled by default and it isn't disabled by griffin in the upstream. 

In my opinion, this feature in its current implementation is harmful because it does not restore default values in case of mismatched hashes for search engine URLs. So for now, it would be better to disable it on every platform to prevent any potential negative user experience..


studies/DisableKeywordTableHashVerification.json5
```diff
@@ -0,0 +1,34 @@
...
+    ],
+    filter: {
+      min_version: '133.*',
+      channel: [
+        'NIGHTLY',
+        'BETA',
+        'RELEASE',
+      ],
+      platform: [
+        'WINDOWS',
```
bsclifton, line 27: Should we remove Windows from this list?

bsclifton (approved):
LGTM! Employees can see conversation at https://bravesoftware.slack.com/archives/C02DT9F5EP5/p1740093092941849

bsclifton:
++ on max version change

bsclifton (approved):

PR #1340, opened 2025-03-25
Increase split view enabled to 50% on desktop

simonhong:


kjozwiak (approved):

PR #1343, opened 2025-03-26
Remove lint-staged.

goodov:
Latest `lint-staged` version requires some tsconfig changes which may need some modification to the setup.

Remove `lint-staged` for now as we can allow running lint on all repo at once without too much trouble.

### DIFF
--- a/src/scripts/lint.ts
+++ b/src/scripts/lint.ts
@@ -4,9 +4,7 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import { program } from '@commander-js/extra-typings';
-import { execSync, spawnSync } from 'child_process';
-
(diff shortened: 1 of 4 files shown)


src/scripts/lint.ts
```diff
@@ -16,23 +14,17 @@ program
       'quality before committing or pushing changes.',
   )
   .option('-a, --all', 'lint all files in the repository')
```
goodov, line 16: the flag is left here intentionally to not break jenkins CI.

atuchin-m: nit: let's change the description to something like "Does nothing, was left for a backward compatibility".

goodov: I'll clean devops repo and remove it now

PR #1349, opened 2025-03-31
Set 100% to split view feature on stable

simonhong:


kjozwiak (approved):

PR #1393, opened 2025-05-08
PartitionVisitedLinkDatabaseWithSelfLinks to 100% on Release

ShivanKaul:
We've had a fairly slow rollout of this. Given that Chrome has it fully enabled, and we've also rolled it out to 100% on Nightly and Beta followed by 25% on Release, seems very safe to fully roll out.

Fix https://github.com/brave/brave-variations/issues/1392


kjozwiak (approved):

PR #1461, opened 2025-08-04
Switch to module: es2022

atuchin-m:


src/test/test_runner.mjs
```diff
@@ -0,0 +1,20 @@
...
+import path from 'node:path';
+
+const targetDir = 'src/';
+let testFiles = [];
+for (const item of readdirSync(targetDir, { recursive: true })) {
+  if (item.endsWith('.test.ts')) {
+    testFiles.push(path.join(targetDir, item));
+  }
+}
+const res = spawnSync('node', ['--import', 'tsx', '--test', ...testFiles], {
```
goodov, line 17: do we need this?

`node --import tsx --test **/*.test.ts` works directly on Windows for me 
```
node --version
v22.13.1
```

atuchin-m: It doesn't work for Ubuntu CI, I've tried ;(
Sounds like `**` is a shell specific thing, we can't relay on it.

An alternative approach is to use `node --import tsx --test`, but we need to pass `--experimental-strip-types` for Node 22. Although, the flag is removed in Node 24, that means we get issues during 22 => 24 transition.

atuchin-m: The good thing that we can remove that file after switching to Node 24 LTS (October 2025).

goodov (approved):

Issue #1485, opened 2025-09-02
BraveNTPBrandedWallpaperStudy


tmancey:
Show new tab takeover ad on first opened tab at launch, then every second tab.

PR #1486, opened 2025-09-02
BraveNTPBrandedWallpaperStudy

tmancey:
Changes NTT so it opens on the 1st tab instead of the 2nd at browser launch, and then on every 2nd tab instead of every 3rd tab thereafter. https://github.com/brave/brave-variations/issues/1485

### DIFF
--- a/studies/BraveNTPBrandedWallpaperStudy.json5
+++ b/studies/BraveNTPBrandedWallpaperStudy.json5
@@ -1,0 +1,44 @@
+[
+  {
+    name: 'BraveNTPBrandedWallpaperStudy',
+    experiment: [
+      {
+        name: 'Enabled',
+        probability_weight: 100,
+        feature_association: {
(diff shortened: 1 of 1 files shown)


aseren (approved):

btlechowski:
Verification passed on 1.81.135 on iPhone 13 Pro Max (iOS 18.5)

Verified NTT is opened on the 1st at browser launch, and then on every 3rd tab

image|image
--|--

btlechowski:
Verification passed on Brave v1.82.161 on Lenovo TB-8506FS (Android 11.0)

Verified NTT is shown every 3rd NTP
image|image|image
--|--|--

On Android there are bugs
https://github.com/brave/brave-browser/issues/47689
https://github.com/brave/brave-browser/issues/49012
which means that the ad will be shown only every 3rd view

Issue #1501, opened 2025-09-15
Disable Tooltips feature on Android to follow upstream


SergeyZhukovsky:
Upstream disables `Tooltips` feature across all channels as it causes crashes and we want to follow as there are crashes as well in GPS. All of them related to Samsung devices and it's unknown currently any STR how to replicate it. https://issues.chromium.org/issues/441235003.

PR #1502, opened 2025-09-15
Disables Tooltips feature on Android across all channels

SergeyZhukovsky:
Resolves: https://github.com/brave/brave-variations/issues/1501


kjozwiak (approved):

PR #1511, opened 2025-09-20
Delete BraveSplitView study

simonhong:
We can delete as it was enabled by default in code also.
And we're going to use SideBySide instead of BraveSplitView.

This doesn't change Brave's current behavior.


kjozwiak (approved):

PR #1517, opened 2025-09-24
Enable UpgradeWhenIdle on Nightly Mac

mherrmann:
The feature only applies to Mac.

Resolves https://github.com/brave/brave-browser/issues/45426.


kjozwiak:
We'll need https://github.com/brave/brave-browser/issues/49604 to be fixed/merged into `master` first before running through the above. As per @mherrmann, STR/Case can be found via https://github.com/brave/brave-core/pull/30753#issue-3342040514.

PR #1520, opened 2025-09-25
Change startup delay to 0s for InProcessBraveAdsServiceStudy

tmancey:
Reduce startup delay from 3s to 0s for InProcessBraveAdsServiceStudy. The delay may have caused us to miss the first new tab takeover opportunity to serve an ad.

Removing it ensures ads initialize immediately, increasing impressions and revenue. This feature has been enabled by default since 1.78.82. This change does not affect iOS, since there is no startup delay there.

Sanity-checked and working as expected. @btlechowski can confirm from the logs that ads initialize faster. There’s still some delay for mojom and related processes, but overall initialization is noticeably quicker.


kjozwiak (approved):

PR #1563, opened 2025-11-25
Disabled search promotion due to startup crash

simonhong:
Promotion button could cause startup crash.
Disabled this study till https://github.com/brave/brave-core/pull/32521 arrives nightly/beta.


kjozwiak (approved):
LGTM 👍Thanks @simonhong. We'll re-enable once we get new `Nightly` & `BETA` with the fix.

PR #1569, opened 2025-12-01
Set `max_version` for `BTMGoogleAuthWorkaround` experiment

bsclifton:
This workaround isn't needed anymore; change has been shipped w/ M141

To see the upstream issue, see https://issues.chromium.org/issues/445561475

For our tracking issue see https://github.com/brave/brave-browser/issues/49300

For original variations PR, see https://github.com/brave/brave-variations/pull/1504

PR #1577, opened 2025-12-05
Revert the recent `BTMGoogleAuthWorkaround` study change

bsclifton:
Seems folks are running into issues without this study enabled.

These are folks on recent versions of Brave, not on older builds.

Some sources
- https://www.reddit.com/r/brave_browser/comments/1pemff1/brave_185111_my_google_account_is_getting/
- https://github.com/brave/brave-browser/issues/50411


bsclifton:
Maximum version was put in place here:
https://github.com/brave/brave-variations/pull/1569

See links in top post for customer reports

Issue #1582, opened 2025-12-11
AB test for Android NTP Idle


SergeyZhukovsky:
Create a study to check 3 variants on Brave Nightly:
A: 4%
B: 4%
C: 4%

PR #1583, opened 2025-12-11
Adds A, B, C variants to AB test Android NTP Idle behaviour

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/1582


kjozwiak (approved):

PR #1622, opened 2026-02-18
Add a max version in BraveAIChatDefaultModelStudy_QwenSonnet_iOS

kylehickinson:
This is when automatic model was landed in iOS when using Leo WebUI.

We unfortunately cant allow this study to still apply if the Leo WebUI feature flag isn't enabled so this version number may need to be adjusted to be where we enable Leo WebUI on iOS


kjozwiak (approved):

Issue #1623, opened 2026-02-19
AB test for Android NTP Idle v2(beta channel)


SergeyZhukovsky:
Create a study to check 2 variants on Brave Beta channel:
A: 50%
B: 50%

PR #1624, opened 2026-02-19
Enables NTP idle AB test on beta channel on 50% for A and 50% for B

SergeyZhukovsky:
Resolves https://github.com/brave/brave-variations/issues/1623


kjozwiak (approved):

PR #1659, opened 2026-04-07
Increases A and B variants to 8 percent on NTP idle study on Android

SergeyZhukovsky:
Resolves: https://github.com/brave/brave-variations/issues/1658


kjozwiak (approved):

PR #1663, opened 2026-04-13
Increase AI Chat WebUI on Release iOS to 50%

kylehickinson:


kjozwiak (approved):

PR #1664, opened 2026-04-13
Enable Brave Shields Content Settings on iOS at 50% in Release

StephenHeaps:
Enabled in release at 25% since March 31st: https://github.com/brave/brave-variations/pull/1648


kjozwiak (approved):

PR #1668, opened 2026-04-14
Rollout AI Chat Conversation API V2 at 50% in Release on Desktop and Android

yrliou:
…Android


kjozwiak (approved):

PR #1674, opened 2026-04-16
Rollout AI Chat Conv API V2 at 100% in Release on Desktop and Android

yrliou:
Metrics still looking good per @porteron, create this so we can rollout at 100% on Thursday night or Friday morning, cc @kjozwiak


kjozwiak (approved):

PR #1683, opened 2026-04-21
Enable Brave Shields Content Settings on iOS at 100% in Release.

StephenHeaps:
Enabled at 50% since April 13th: https://github.com/brave/brave-variations/pull/1664#event-24463466553


LaurenWags (approved):
Uplift to production approved after conferring with @brave/uplift-approvers 👍🏻

PR #1656, opened 2026-04-03
Rollout `AIChatCodeExecutionTool` to 100% beta, 25% release

DJAndries:
Already tested in https://github.com/brave/brave-variations/pull/1576

PR #1684, opened 2026-04-22
Increase `AIChatCodeExecutionTool` rollout to 50%

DJAndries:
Previous rollout: https://github.com/brave/brave-variations/pull/1656


LaurenWags (approved):
Uplift to production is approved after conferring with @brave/uplift-approvers 👍🏻

PR #1713, opened 2026-05-19
[documentation] Improve doc for min_version and max_version

### LINKED ISSUE #1712: [Documentation] Update comment for min_version and max_version in study.proto
We currently have the following comment for `min_version` and `max_version` in the [study.proto](https://github.com/brave/brave-variations/blob/main/src/proto/study.proto#L313-L324). 
```

// The minimum Chrome version for this study, allowing a trailing '*'
// character for pattern matching. Inclusive. (To check for a match, iterate
// over each component checking >= until a * or end of string is reached.)
// Optional - if not specified, there is no minimum version.
// Ex: "17.0.963.46", "17.0.963.*", "17.*"
optional string min_version = 2;

// The maximum Chrome version for this study; same formatting as
// |min_version| above. Inclusive. (To check for a match, iterate over each
// component checking <= until a * or end of string is reached.)
// Optional - if not specified, there is no maximum version.
// Ex: "19.*"
optional string max_version = 3;
```

The comment says to put the minimum Chrome version, w/o specifying anything about Brave version which is not correct. From the error log [here](https://github.com/brave/brave-variations/actions/runs/26057565060/job/76609347517), it states to have the following format: `CHROMIUM_MAJOR.BRAVE_MAJOR.BRAVE_MINOR.BRAVE_BUILD `. We should update the comment to reflect it in the proto file.

r0hit0303:
Resolves https://github.com/brave/brave-variations/issues/1712


goodov (approved):

antonok-edm (approved):

r0hit0303:
Thank you for the review!

PR #1716, opened 2026-05-21
Enable iOS Wallet WebUI at 100% in Release

nuo-xu:
was at 25% since May 12th. we will ramp up faster to set it to 100% since iOS wallet user is fairly low.


kjozwiak (approved):

PR #1727, opened 2026-05-28
Bump conversation API v2 stop version to start with CR148

yrliou:
This would mean we stop the study after any 1.91.x with CR148.


kjozwiak (approved):

PR #1740, opened 2026-06-08
Drop react-router-dom dep

atuchin-m:
For https://github.com/brave/brave-variations/security/dependabot/91
For https://github.com/brave/brave-variations/security/dependabot/88


goodov (approved):
nice 👍

PR #1741, opened 2026-06-10
Enforce pnpm as a package manager to prevent npm install to work.

goodov:
Use `devEngines` field as a hard stop to prevent `npm install` to work. Without this `npm install` will be happy to roll whatever is written in `package.json` with all most recent sub-dependencies, because there's no lock file for npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,18 @@
     "webpack-cli": "7.0.2",
     "webpack-dev-server": "5.2.4"
   },
-  "engines": {
-    "node": ">=22.0.0"
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": ">=22.0.0",
+      "onFail": "error"
+    },
+    "packageManager": {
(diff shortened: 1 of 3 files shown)


atuchin-m (approved):

package.json
```diff
@@ -45,10 +45,18 @@
...
-    "node": ">=22.0.0"
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": ">=22.0.0",
+      "onFail": "error"
+    },
+    "packageManager": {
+      "name": "pnpm",
+      "version": ">=10.29.2",
```
atuchin-m, line 56: the range here works only for >= 11:
```
devEngines.packageManager now supports version ranges;
```
https://pnpm.io/blog/releases/11.0

goodov: as long as it blocks npm install, that's good enough for now.

PR #1743, opened 2026-06-10
Revert "Enforce pnpm as a package manager to prevent npm install to work."

atuchin-m:
Reverts brave/brave-variations#1741
We need to update pnpm to 11 until we can do it.

```
18:02:03    stdout         : 'Invalid package manager specification in package.json (pnpm@>=10.29.2); expected a semver version\n'
```

https://bravesoftware.slack.com/archives/C06NFG8NAR2/p1781100127481789

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,18 +45,10 @@
     "webpack-cli": "7.0.2",
     "webpack-dev-server": "5.2.4"
   },
-  "devEngines": {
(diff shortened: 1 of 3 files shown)


goodov:
fixed with https://github.com/brave/brave-variations/pull/1745

PR #1744, opened 2026-06-11
Enable `load_after_restore` for `BraveSearchBackupResults`

DJAndries:
Rolling out the change described in https://github.com/brave/internal/issues/1762. QA is optional


kjozwiak:
Waiting till https://github.com/brave/brave-browser/issues/56345 goes out via the next `1.91.x` minor release to see if that improved things before merging the above into `production`. Waiting on the minor `CR149` which we'll probably get either today/tomorrow.

PR #1746, opened 2026-06-11
Revert pnpm enforcement

goodov:
perf agent uses corepack right now, so it wants packageManager field to be the exact version. We need to fix the corepack restriction first.


goodov:
corepack is removed from perf agent runs.

PR #1750, opened 2026-06-15
Bump few deps.

goodov:
- Resolves https://github.com/brave/brave-variations/security/dependabot/96
- Resolves https://github.com/brave/brave-variations/security/dependabot/94


goodov:
ptal @tmancey

atuchin-m (approved):

PR #1756, opened 2026-06-23
[iOS] Disable `NetworkConnectionIntegrityStudy` on all channels

StephenHeaps:
Currently only enabled via Griffin on Nightly.
Disabling temporarily due to page load issues on iOS 27 betas, `brave://flags` won't load so it can only be disabled from Developer Options / QA Menu switches.


StephenHeaps:
If preference is to delete the study, please let me know and I'll update this PR.

kjozwiak (approved):

PR #1766, opened 2026-07-03
Migrate to pnpm 11

atuchin-m:


package.json
```diff
@@ -52,7 +52,7 @@
     },
     "packageManager": {
       "name": "pnpm",
-      "version": ">=10.29.2",
+      "version": ">=11.7.0",
```
goodov: 11.9.0 and adjust node to be >=24

also please add `minimumReleaseAge: 10080` to pnpm-workspace and re-generate lock file.

atuchin-m: 1. done.
2. `minimumReleaseAge` is already added.

Issue #1769, opened 2026-07-06
Disable PopoverHintNewBehavior (ChatGPT web issues)


atuchin-m:
The feature was disabled [here](https://chromium-review.googlesource.com/c/chromium/src/+/8012419) and the fix was uplifted to [cr150](https://chromium-review.googlesource.com/c/chromium/src/+/8027705) & [cr151](https://chromium-review.googlesource.com/c/chromium/src/+/8022722).
For cr151 the fix is available from 150.0.7871.67 (isn’t created yet)

The feature is the root cause of ChatGPT issues: https://github.com/brave/brave-browser/issues/56900

Slack thread: https://bravesoftware.slack.com/archives/C05S50MFHPE/p1782954756658209

atuchin-m:
closed in favor of releasing an update with chromium 150.0.7871.101

PR #1770, opened 2026-07-06
Disable PopoverHintNewBehavior (stable 5%)

atuchin-m:
For https://github.com/brave/brave-variations/issues/1769
For https://github.com/brave/brave-browser/issues/56900

Nightly & Beta: 100%
Stable: 5%

### DIFF
--- a/studies/PopoverHintNewBehavior.json5
+++ b/studies/PopoverHintNewBehavior.json5
@@ -1,0 +1,61 @@
+[
+  {
+    name: 'PopoverHintNewBehavior',
+    experiment: [
+      {
+        name: 'Disabled_EmergencyKillSwitch',
+        probability_weight: 100,
+        feature_association: {
+          disable_feature: [
+            'PopoverHintNewBehavior',
+          ],
+        },
+      },
+    ],
+    filter: {
+      min_version: '150.*',
+      max_version: '152.*',
+      channel: [
+        'NIGHTLY',
+        'BETA',
+      ],
+      platform: [
+        'WINDOWS',
+        'MAC',
+        'LINUX',
+        'ANDROID',
+      ],
+    },
+  },
+  {
+    name: 'PopoverHintNewBehavior',
+    experiment: [
+      {
+        name: 'Disabled_EmergencyKillSwitch',
+        probability_weight: 5,
+        feature_association: {
+          disable_feature: [
+            'PopoverHintNewBehavior',
+          ],
+        },
+      },
+      {
+        name: 'Default',
+        probability_weight: 95,
+      },
+    ],
+    filter: {
+      min_version: '150.*',
+      max_version: '152.*',
+      channel: [
+        'RELEASE',
+      ],
+      platform: [
+        'WINDOWS',
+        'MAC',
+        'LINUX',
+        'ANDROID',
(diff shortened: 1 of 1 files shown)


atuchin-m:
Tested on `v1.94.37 Chromium: 150.0.7871.63 (Official Build) nightly (arm64) `:
voice input works on https://chat.openai.com/ 

image

nazedaque:
I tested `--variations-pr=1770`, but the issue still reproduces for me (Stable).

Test procedure:

* Launched Brave (Stable) with `--variations-pr=1770`
* Waited around 10 seconds
* Fully restarted Brave
* Verified the test seed in `brave://version`
* Active Variation starts with: `pull/1770@f36c5139dd7b69917c94bc017bcf4e70a6c8fbb1`

Result: the ChatGPT web UI still fails with:

Content failed to load

**Important difference from the original Voice Mode reports: I am not using the microphone, dictation, or Voice Mode at all. This happens during normal text-only ChatGPT usage.**

I also see `?mweb_fallback=1` appended to the URL after the failure, but my interpretation is that this may be a secondary fallback triggered after I refresh the page while ChatGPT is still processing, rather than the root cause.

Environment:

* Brave version: 1.92.134 
* Chromium version: 150.0.7871.63
* OS: Windows 11 Pro 25H2
* Input mode: text only, no microphone, no dictation, no Voice Mode
* Logout/login: tested, did not fix the issue
* Shields off: tested
* Extensions disabled: tested
* Firefox: works better and does not reproduce the same failure in the same way

So disabling `PopoverHintNewBehavior` via this variation does not appear to fully resolve my case, at least for normal text-only ChatGPT usage.

nazedaque:
This time I tested Brave **Beta** with `--variations-pr=1770`.

Result: the ChatGPT web UI is now much more stable. I have not reproduced the previous failure:

Content failed to load

This was tested with normal text-only ChatGPT usage, without microphone, dictation, or Voice Mode.

**However, I noticed a possible side effect: the dynamic status text that normally appears while a GPT Thinking response is being generated no longer appears during generation.**

Normally, while the model is thinking, ChatGPT shows a small dynamic text/status element that can be clicked to open a right-side panel. With this variation active, that element does not appear for me anymore.

So from my test:

* Brave Beta + `--variations-pr=1770` appears to fix the main ChatGPT UI crash.
* But the Thinking/status popover or right-side panel trigger may no longer be displayed during generation.
* This may be an expected consequence of disabling `PopoverHintNewBehavior`, but I am reporting it in case it is an unintended UI regression.

Environment:

* Brave channel: Beta
* Brave version: 1.93.110
* Chromium version: 150.0.7871.63
* OS: Windows 11 Pro 25H2
* Variation: `--variations-pr=1770`
* Active seed confirmed in `brave://version`: yes
* ChatGPT usage mode: text only, no microphone, no dictation, no Voice Mode

atuchin-m:
@nazedaque the chance you get the fix on Stable is only 5% for now. We will increase it into the next PRs.
To forcefully disable the problematic feature use `--disable-features=PopoverHintNewBehavior`

nazedaque:
I can confirm that forcing the feature off resolves the issue for me on Brave Stable.

PR #1788, opened 2026-07-30
[BraveRequestInfoUniquePtr] Add a max version as default value was changed

bsclifton:
See https://github.com/brave/brave-core/pull/38453 for where default value was changed

I think it makes sense to let that ride the trains (or uplift if needed) and more towards deprecating the Griffin flag


kjozwiak (approved):
Uplift into `production` approved 👍

PR #1802, opened 2026-08-11
Roll out Brave Ads oblivious HTTP to 100% on Release

tmancey:
Bumps Release from 50% to 100% and merges the Release study entry
into the existing Nightly/Beta entry, since all channels now share
the same rollout percentage.


kjozwiak (approved):

Issue #1804, opened 2026-08-12
Track outdated studies


atuchin-m:
1. Add a script and a CI action to track outdated studies.
2. Set the max_version for studies are already RELEASE 100%. This is likely imply b-c changes if the feature isn't enabled by default.
3. Abandon hanging studies (i.e. enable for Nightly a year ago and never released)
4. Introduce a policy of the study lifecycle.

PR #1805, opened 2026-08-12
Add a script and an action to track outdated studies

atuchin-m:
For https://github.com/brave/brave-variations/issues/1804

A study is considered outdated iff a) there is no `max_version` and b) it was modified 3+ months ago. 
The new gh action is currently disabled (until we fix the most studies).

the current output (84 studies out of 130):
https://gist.github.com/atuchin-m/a634755c35a7d854e730f9469000897e

An example of mkrdwn (slack format) output:
image

### DIFF
--- a/src/seed_tools/commands/find_outdated_studies.ts
+++ b/src/seed_tools/commands/find_outdated_studies.ts
@@ -89,8 +89,8 @@ async function findOutdatedStudies(
     const content = await fs.promises.readFile(filePath, 'utf8');
     const { studies, errors } = parseStudyFile(filePath, content);
     if (errors.length > 0) {
-      console.error(`Skipping ${filePath}:\n${errors.join('\n')}`);
-      continue;
+      console.error(`Error parsing ${filePath}:\n${errors.join('\n')}`);
+      process.exit(1);
     }
 
     if (studies.every((study) => study.filter?.max_version)) {
@@ -123,7 +123,8 @@ function getLastCommit(filePath: string): {
     { encoding: 'utf8' },
   ).trim();
   if (!output) {
-    return { commit: '', date: fs.statSync(filePath).mtime };
+    console.error(`Failed to get last commit for ${filePath}`);
+    process.exit(1);
   }
   const [commit, date] = output.split('\n');
   return { commit, date: new Date(date) };
@@ -162,7 +163,7 @@ function findPrAuthor(commit: string): string {
   } catch (e) {
     const message = e instanceof Error ? e.message : String(e);
     console.error(`Failed to look up PR for ${commit}: ${message}`);
-    return 'unknown';
+    process.exit(1);
   }
 }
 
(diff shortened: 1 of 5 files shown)


tmancey (approved):
LGTM

src/seed_tools/commands/find_outdated_studies.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  const outdated: OutdatedStudy[] = [];
+
+  for (const file of files) {
+    const filePath = path.join(studiesDir, file);
+    const content = await fs.promises.readFile(filePath, 'utf8');
+    const { studies, errors } = parseStudyFile(filePath, content);
+    if (errors.length > 0) {
+      console.error(`Skipping ${filePath}:\n${errors.join('\n')}`);
+      continue;
```
goodov: I think this should be process.exit(1)

src/seed_tools/commands/find_outdated_studies.ts
```diff
@@ -0,0 +1,202 @@
...
+  commit: string;
+  date: Date;
+} {
+  const output = execFileSync(
+    'git',
+    ['log', '-1', '--format=%H%n%cI', '--', filePath],
+    { encoding: 'utf8' },
+  ).trim();
+  if (!output) {
+    return { commit: '', date: fs.statSync(filePath).mtime };
```
goodov: is this the right fallback? I don't understand when it might happen and why we're not just failing here.

src/seed_tools/commands/find_outdated_studies.ts
```diff
@@ -0,0 +1,202 @@
...
+  const [commit, date] = output.split('\n');
+  return { commit, date: new Date(date) };
+}
+
+function findPrAuthor(commit: string): string {
+  if (!commit) {
+    return 'unknown';
+  }
+  try {
+    const output = execFileSync(
```
goodov, line 138: what if it fail, the output will be empty or the tool will terminate? `gh` call can fail randomly, so this should work right, i.e. we don't want show 'unknown' for all outdated studies on this occasion.

atuchin-m: Frankly, tracking the commit/PR isn't a critical thing, it's extra info that should help to identify the owner.
Let's use ` process.exit(1)` anyway.

src/seed_tools/commands/find_outdated_studies.ts
atuchin-m: It's for untracked files. Let's replace to `process.exit(1)`

src/seed_tools/commands/find_outdated_studies.ts
atuchin-m: Fixed.

src/seed_tools/commands/find_outdated_studies.ts
atuchin-m: Done.

src/seed_tools/commands/find_outdated_studies.ts
atuchin-m: Done.

goodov (approved):

Issue #1817, opened 2026-08-13
Epic ticket for stuck studies (enabled for some users, but never reach Release 100%)


atuchin-m:
The current list
```
BraveAdblockExperimentalListDefaultStudy        28 Oct 2024     goodov
BraveLocalhostAccessPermissionStudy     28 Oct 2024     goodov
BraveRequestOTRTabRolloutStudy  28 Oct 2024     goodov
BraveSearchPromotionBannerStudyOnStable 28 Oct 2024     goodov
BraveWalletAnkrBalancesEnabled  28 Oct 2024     goodov
BraveWebcompatExceptionsServiceStudy    28 Oct 2024     goodov
CosmeticFilteringChildFramesStudy       28 Oct 2024     goodov
DefaultBraveCommandsStudy       28 Oct 2024     goodov
DefaultPlaylistStudy    28 Oct 2024     goodov
ModuleFileNamePatchStudy        28 Oct 2024     goodov
ModuleFileNamePatchStudyRelease 28 Oct 2024     goodov
NewiOSPlaylistUIStudy   28 Oct 2024     goodov
BraveSearchPromotionBannerStudy 17 Jan 2025     simonhong (Simon Hong)
HangWatcher     25 Feb 2025     atuchin-m (Mikhail)
BraveRewardsPlatformCreatorDetectionStudy       18 Apr 2025     deeppandya (Deep Pandya)
BraveCopyCleanLinkFromJsStudy   23 Jul 2025     boocmp (Pavel Beloborodov)
BraveAdblockShowHiddenComponentsStudy   6 Aug 2025      antonok-edm (Anton Lazarev)
BraveFallbackDoHStudy   16 Sept 2025    arthuredelstein (Arthur Edelstein)
BraveWebViewRoundedCornersStudy 3 Dec 2025      zenparsing (Kevin Smith)
BraveSearchPromotionButtonStudy 15 Dec 2025     simonhong (Simon Hong)
UpgradeWhenIdleStudy    22 Jan 2026     mherrmann (Michael Herrmann)
ShowBraveShieldsInPageInfoStudy 30 Jan 2026     zenparsing (Kevin Smith)
UseOmaha4Study  24 Mar 2026     mherrmann (Michael Herrmann)
ForceContextMenuOnShiftRightClickStudy  23 Apr 2026     bsclifton (Brian Clifton)
HistoryEmbeddingsParamsStudy    27 Apr 2026     darkdh (Anthony Tseng)
```

PR #1840, opened 2026-08-26
Remove HangWatcher study

atuchin-m:
For https://github.com/brave/brave-variations/issues/1817
It's a dev-concern study that reports hangs as crashes.
The Chromium default  in the code now uses the same values: https://source.chromium.org/chromium/chromium/src/+/main:base/threading/hang_watcher.cc;l=257-269?q=EnableHangWatcher
We don't need to enforce it via Griffin.


kjozwiak (approved):

PR #1844, opened 2026-08-26
Remove AndroidBottomToolbar study as stale

### LINKED ISSUE #1843: Stale study: AndroidBottomToolbar.json5 (flag renamed to AndroidBottomBar)
The Griffin study still enables `AndroidBottomToolbar` at 100% on Android (all channels, `min_version: 139.*`):

https://github.com/brave/brave-variations/blob/main/studies/AndroidBottomToolbar.json5#L10

Last modified: 19 Aug 2025, samartnik. Rollout: https://github.com/brave/brave-variations/issues/1479

That flag **no longer exists**. Chromium renamed it to `AndroidBottomBar`. In latest stable (v1.93.138 / Chromium 151.0.7922.173) there is no `AndroidBottomToolbar` `BASE_FEATURE`; the replacement is still DISABLED:

https://github.com/chromium/chromium/blob/151.0.7922.173/chrome/browser/flags/android/chrome_feature_list.cc#L599

Java CachedFlag default is also false:
https://github.com/chromium/chromium/blob/151.0.7922.173/chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureList.java#L794-L795

Because Griffin targets the old name, the study is a no-op on current clients.

This experiment should be ended by one of:

* Delete `AndroidBottomToolbar.json5`— the flag is gone, so the study does nothing.
* Rename the study to `AndroidBottomBar` and keep rolling it out — only if product still wants the bottom bar on at 100%. That would also need a matching code default (the C++/Java flag is still DISABLED).

Do not leave the current study as-is: it ships a deleted flag name.

samartnik:
Resolves https://github.com/brave/brave-variations/issues/1843


kjozwiak (approved):

PR #1859, opened 2026-09-02
chore(deps): Use separate Renovate group for pnpm

mschfh:


package.json
```diff
@@ -56,6 +56,7 @@
       "onFail": "error"
     }
   },
+  "packageManager": "pnpm@11.9.0+sha256.2b567aa66026238078ac2e0a33bec3febd60e962987aac697456f3180819b287",
```
goodov, line 59: won't this conflict with `devEngines`? I remember there was something not right when both fields existed.

goodov (approved):

PR #1870, opened 2026-09-09
Set max_version for DisableChromiumGlicExtension

atuchin-m:
`DisableChromiumGlicExtension` study was added here: https://github.com/brave/brave-core/pull/38843
Hardcoded in https://github.com/brave/brave-core/pull/38843


kjozwiak (approved):

Issue #1874, opened 2026-09-10
Stale study: BraveDayZeroStudyAndroid (feature: BraveDayZeroExperiment)


atuchin-m:
[`BraveDayZeroStudyAndroid.json5`](https://github.com/brave/brave-variations/blob/main/studies/BraveDayZeroStudyAndroid.json5) never reached 100% on Nightly+Beta+Release, but the feature is still enabled for some users (Android Release **100%**, `variant: B`, `min_version: 146.*`). It also enables `BraveFreshNtpAfterIdleExperiment`.

**Last modified:** 2 Jun 2026, [#1731](https://github.com/brave/brave-variations/pull/1731), SergeyZhukovsky (Serg)

Root issue: https://github.com/brave/brave-variations/issues/1817

This experiment should be ended by one of:

* **Abandon the rollout** — remove the config, or add `max_version`.
* **Keep the current behavior in code** — hard-code the intended channel/platform/language (etc.) targeting, then add `max_version`.
* **Continue the rollout** — only if this is still important and was missed.

Same feature is also enabled at 6% on Windows Release by [`BraveDayZeroStudyDesktop.json5`](https://github.com/brave/brave-variations/blob/main/studies/BraveDayZeroStudyDesktop.json5).

SergeyZhukovsky:
I pushed `max_version` for Android fix, desktop study is really unrelated and probably is still going.

PR #1877, opened 2026-09-10
Sets max_version for BraveFreshNtpAfterIdleExperiment feature as we enabled it by default

SergeyZhukovsky:
The feature was enabled by default in code here https://github.com/brave/brave-core/pull/39010
Resolves: http://github.com/brave/brave-variations/issues/1874


atuchin-m (approved):

kjozwiak (approved):

PR #1880, opened 2026-09-11
[Polkadot] [Wallet] increase Polkadot rollout percentage to 25%

cmazakas:
Increase rollout percentage of Polkadot to 25% on the release channel.


kjozwiak (approved):

